PR #2, opened 2020-05-20
Simplify instructions about out-of-tree build

jserv:


PProvost:
Interesting suggestion. Both sequences result in the same outcome. I'm curious why you prefer the long-winded version?

jserv:
> Interesting suggestion. Both sequences result in the same outcome. I'm curious why you prefer the long-winded version?

My intention was not to confuse users who are not familiar with CMake, so that they can simply copy-n-paste the instructions and ensure static library (maybe more objects are generated later) appears in the `build` directory without problems.

PProvost:
Fair enough. Since the two command sequences are identical in outcome, I'm going to leave the more terse form in place for now. If people get confused by it, we can reconsider.

I'm going to close this PR. Feel free to open an Issue if you want to continue this discuss. Thanks for your time on this.

PR #18, opened 2020-07-14
Updated tx_api.h

ameya7295:
Added a new variable INT tx_errno, This field is needed by nx_bsd.c : 8271


goldscott:
Hi @ameya7295 
We only want tx_errno to be defined when using the BSD wrapper. We have extensions in the TX_THREAD struct to add the errno. I also saw your suggestion in https://github.com/azure-rtos/threadx/issues/17 - we'll discuss internally and try to make some improvements.

goldscott:
We are going to keep using bsd_errno rather than tx_errno.

PR #25, opened 2020-07-28
Fix Windows build by copying include directories from debug to release.

BillyONeal:


goldscott:
Hi @BillyONeal - thanks for your fix! Looks like we had our old directory structure settings for the release build. We'll get this fixed internally and then update github during our next patch release. Thanks!

goldscott:
Hi @BillyONeal - most IDEs automatically create two configuration: Debug and Release. Supporting both means twice the amount of effort on our part. Since the IDEs use the Debug configuration as default, I see the Debug configuration as the Default configuration - since 'Debug' is just the name of the configuration, it might as well be 'Default' IMO. Users are free to create their own configurations (e.g. Release).

BillyONeal:
@goldscott Of course if the user creates new configurations they are on their own, but these configurations are shipped in your sources. Deleting one of them would also be OK but that would have been a bigger change than fixing the one problem, that being the wrong include path.

(To be clear, I discovered this problem trying to add threadx to vcpkg; vcpkg builds both debug and release flavors and the build fails because the release flavor here was broken)

BillyONeal:
I see you closed this; does that mean you're going to replay the changes internally or you're just leaving a broken build script check in? I'm confused.

goldscott:
Hi @BillyONeal - release has been fixed!

BillyONeal:
Sounds good, thanks!

PR #57, opened 2020-12-04
New port files for Cortex-M (C source, 

kjassmann-renesas:
New port files based on ra/microsoft/azure-rtos/threadx/ports/cortex_m33/iar and Renesas RA port.

Major updates from CM33/IAR port include:
 * No assembly files. The port is mostly C, with inline assembly used where necessary.
 * Uses CMSIS core functions where possible.
 * Uses Arm TrustZone RTOS Context Management functions (https://arm-software.github.io/CMSIS_5/Core/html/group__context__trustzone__functions.html).
 * Supports Cortex-M33, Cortex-M23, and Cortex-M4 in a single code base.
 * Supports GCC (9-2019-q4-major), IAR (8.50.7), and Arm Compiler 6 (6.14) in a single code base.

### DIFF
--- a/ports/cortex_m_cmsis/example_build/tx_cmsis_sample.h
+++ b/ports/cortex_m_cmsis/example_build/tx_cmsis_sample.h
@@ -1,0 +1,29 @@
+/**************************************************************************/
+/*                                                                        */
+/*       Copyright (c) Microsoft Corporation. All rights reserved.        */
+/*                                                                        */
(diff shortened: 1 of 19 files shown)


goldscott:
Hi @kjassmann-renesas - I'm going to close this because you've talked about it in private with our team. I think in the future we will try to migrate from assembly to C and try to support compilers with a single port. We like what you've done!

PR #81, opened 2021-02-24
Add thread local storage option

slide:
This adds the possibility of having thread local storage items in the thread context structure. By default, the number of slots is 0, so no extra space is taken up by the current configs. If the tx_port.h sets TX_THREAD_LOCAL_STORAGE_SLOTS to a number > 0, then that number of slots will be available for use with the functions provided. This is an opt-in feature.


slide:
I setup a similar build on my Jenkins instance and all four of the failing items here are building correctly. There seems to be a possible issue on the build setup?

goldscott:
Hi @slide, I'm not sure if we want to add these new APIs to ThreadX. Users are free to use the TX_THREAD_USER_EXTENSION and put whatever they want there.

slide:
Ok. I wasn't aware of that, I'll check into it.

goldscott:
@slide - there are a lot of #defines that aren't documented in the manual - we are working on updating and making these optional compile-time features more clear. Cheers!

PR #85, opened 2021-03-05
Adds space in TX_LOW_POWER and TX_ENABLEWFI defines

jmmb13:
In schedulers adds space in IF:DEF:TX_LOWER_POWER to remove error 'Unkown opcode IF:DEF:TX_LOW_POWER , expecting opcode or Macro'


goldscott:
Ugh, AC5 is such a pain! I'll fix this internally and it will be available next patch. We literally just released 6.1.5 a few minutes ago.

goldscott:
I just fixed this for M0, M3, M4, M7.

PR #105, opened 2021-05-25
common: fix wrong down casting in _tx_thread_create

sagi-h:
casting new_stack_start to ULONG can result in the wrong start address when running on 64-bit 
machines (e.g. gnu/linux) as it zeroes the higher address bits. The correct type is ALIGN_TYPE, since 
both new_stack_start and updated_stack_start are ALIGN_TYPE the casting can be simply removed 
and this change should have no effect on 32-bit machines.

### DIFF
--- a/common/src/tx_thread_create.c
+++ b/common/src/tx_thread_create.c
@@ -121,7 +121,7 @@ ALIGN_TYPE              updated_stack_start;
 
     /* Ensure the starting stack address is evenly aligned.  */
     new_stack_start =  TX_POINTER_TO_ALIGN_TYPE_CONVERT(stack_start);
-    updated_stack_start =  ((((ULONG) new_stack_start) + ((sizeof(ULONG)) - ((ULONG) 1)) ) & (~((sizeof(ULONG)) - ((ULONG) 1))));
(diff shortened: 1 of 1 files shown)


goldscott:
Hi @sagi-h - thanks for finding this! Sorry for the delayed response - I've was on vacation for a few weeks. We are going to review this internally and see if there are any other mis-casts. This should be fixed in the next ThreadX release (hopefully in August).

sagi-h:
> Hi @sagi-h - thanks for finding this! Sorry for the delayed response - I've was on vacation for a few weeks. We are going to review this internally and see if there are any other mis-casts. This should be fixed in the next ThreadX release (hopefully in Augu

Thanks Scott, while you are at it, could you also revisit the related typedef for LONG/ULONG on the linux/GNU port:
```
#if __x86_64__
typedef int                                     LONG;
typedef unsigned int                            ULONG;
```

goldscott:
Hi @sagi-h, 

For our 64-bit Linux port, to keep ThreadX consistent, we need to make sure size of LONG and ULONG is 32 bits. But size of long is 64 bits in Linux. That is the reason LONG is mapped to int rather than long.

sagi-h:
Fixed in 6.1.8

PR #126, opened 2021-07-05
Enable Arm Cortex-M33 Support

microbuilder:
This commit enables support for the Arm Cortex-M33, using the GNU toolchain. By default, HW DSP or FP support has been disabled, and compilation takes place with `-mfloat-abi=softfp`.

It also address an issue with `tx_thread_schedule.S` where floating point instructions were being hard-coded into the assembly code, causing a fault on devices that don't have HW FPU support, such as the MPS3 AN524 Core 0, etc.

Signed-off-by: Kevin Townsend <kevin.townsend@linaro.org>
Signed-off-by: Jimmy Brisson <jimmy.brisson@linaro.org>


goldscott:
Hi @microbuilder - thanks for bringing this up! We'll fix this internally and it'll be in our next release.

goldscott:
Update: arm-none-eabi.cmake  and tx_thread_schedule.S have been fixed internally and will be available in our next release, scheduled for August 2. At this time, we are not going to add/maintain new cmake files.

PR #185, opened 2022-04-20
Fix Typo: Auzre -> Azure

skade:


goldscott:
Thanks for finding this typo! We'll fix it internally and it will be changed in our 6.1.12 release.

Issue #172, opened 2022-03-10
low power on ARM: LR should be saved in _tx_thread_schedule


foldl:
Task this as an example, after `BL tx_low_power_exit`, `LR` is overwritten. Later, when a new task is ready to run, `BX lr` would not work.

https://github.com/azure-rtos/threadx/blob/master/ports/cortex_m0/keil/src/tx_thread_schedule.s#L264

```asm
    IF :DEF:TX_LOW_POWER
    PUSH    {r0-r3}
    BL      tx_low_power_exit                       ; Exit low power mode
    POP     {r0-r3}
    ENDIF
```

goldscott:
Yep, you're right. We'll fix this in the next release.

goldscott:
It doesn't matter what LR is here. When a thread is scheduled, it's LR will be restored.

foldl:
Yes, `LR` is restored.

I would like to provide more information on why I have created this issue.

In `TX_LOW_POWER_USER_ENTER`, MCU is powered off while RAM is retention-ed. After `tx_low_power_next_expiration`,
MCU is powered on again, I try to resume the execution:

```c
void low_power_user_enter()
{
    power off;
restore_here:  // after MCU powered on again (Thread mode), jump to restore_here
    restore context;
}
```

This works for RTOSes that have a dedicated idle task. While in ThreadX, `low_power_user_enter` is called in PendSV, it needs
extra tinkering to "resume" the PendSV ISR.

foldl:
Create an idle task and my solution works. Suggest adding documentation on this, warnings and tips.

foldl:
To resume/return to the PendSV ISR, I think we can:

1. After MCU powered on again from low power mode, set up a flag and trigger PendSV;
2. In PendSV, when the flag is set, jump to `restore_here`.

This solution is more complex than using an idle task, so I gave it up and created a dedicated idle task to do the work, :)

foldl:
Here is what I am doing before creating a dedicated idle task (ARM Cortex-M is assumed):

```c
reset_handler()
{
    if (wakeup from low power) jump to restore_here;
    ...
}

void low_power_user_enter()
{
    power off;
restore_here:  // after MCU powered on again (Thread mode), jump to restore_here
    restore context;
}
```

The concept here is to let caller of `low_power_user_enter` not feel anything about the low power procedure. 

[...]

PR #201, opened 2022-07-13
add info on low power based on the idle thread

foldl:
It's better to give some information on low power based on the idle thread.
An example (template) of the idle thread is given.

There are other RTOSes that have an idle task as default. For developers who are
porting projects, with this information available, I think issues like #172 can be avoided.

### DIFF
--- a/utility/low_power/low_power.md
+++ b/utility/low_power/low_power.md
@@ -213,7 +213,7 @@ ULONG tx_timer_get_next(ULONG *next_timer_tick_ptr);
 
 ### Description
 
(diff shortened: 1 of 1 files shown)


goldscott:
Hi @foldl - we have updated the low_power.md documentation and it will be available in the next release, scheduled for the end of July. We do make the suggestion that an idle thread is one possible implementation, but we do not have an example of an idle thread.

goldscott:
We updated the documentation in the 6.1.12 release: https://github.com/azure-rtos/threadx/blob/master/utility/low_power/low_power.md

PR #202, opened 2022-07-21
Support for CMake compilation for Cortex M33

asierllano:
The CMakeLists.txt file has been included to support compilation
with CMake for Cortex M33. It has been entirely based on Cortex M4
compilation system. It has been tested and worked for STM32U575
device.

Signed-off-by: Asier Llano <allano@hubbell.com>


goldscott:
Hi @asierllano - we actually added CMake support to the M33 internally a few weeks ago and it will be released at the end of July. Similar files are available now in the M55 and M85 ports. Here's a link: https://github.com/azure-rtos/threadx/tree/master/ports/cortex_m55
I'll close your PR as this update has already been done. Thanks for your contribution!

PR #228, opened 2023-02-17
fix a comment about GIC

hodcarrier:
[1] from a cortex-a9 soc's manual:
0100—01FFh Interrupt controller interfaces
1000—1FFFh Interrupt distributor
[2] arm_cortexa9_trm_100511_0401_10_en.pdf
4.3.25 Configuration Base Address Register


goldscott:
This is code from ARM that we supply as part of an example. We do not maintain this code from ARM.

PR #256, opened 2023-04-28
Add random number stack filling option

xiuwencai:
Add an option to use random number to fill stacks. This option can be enabled only when stack checking is enabled.


common/src/tx_initialize_kernel_enter.c
```diff
@@ -133,6 +137,9 @@ VOID  _tx_initialize_kernel_enter(VOID)
        later used to represent interrupt nesting.  */
     _tx_thread_system_state =  TX_INITIALIZE_IN_PROGRESS;
 
+    /* Optional random number generator initialization.  */
+    TX_INITIALIZE_RANDOM_GENERATOR_INITIALIZATION
```
bo-ms, line 141: Could you explain why we add this feature? and could we let user initialize the random number generator in application code before calling _tx_initialize kernel_enter()?

PR #264, opened 2023-05-19
Export _tx_handler_svc_unrecognized as weak symbol.

xiuwencai:
Export _tx_handler_svc_unrecognized as weak symbol, applications can implement customized function to handle other SVCs.


TiejunMS (changes requested):
Could you update function headers?

TiejunMS (approved):

PR #275, opened 2023-06-16
Unify ThreadX and SMP for ARMv8-A.

TiejunMS:
## PR checklist

### DIFF
--- a/ports/cortex_a35/gnu/example_build/tx/.cproject
+++ b/ports/cortex_a35/gnu/example_build/tx/.cproject
@@ -39,7 +39,7 @@
                             							
                             <tool id="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.var.gcc-8.3.0-aarch64-elf.1755052156" name="GCC C Compiler 8.3.0 [aarch64-elf]" superClass="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.var.gcc-8.3.0-aarch64-elf">
                                 								
-                                <option id="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.option.cpu.281590127" name="CPU (-mcpu)" superClass="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.option.cpu" useByScannerDiscovery="false" value="cortex-a35" valueType="string"/>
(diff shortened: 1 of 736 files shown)


hodcarrier:
hello, ARMv8-A is hard to type... other projects refers it as armv8-a, gcc's compile option is -march=armv8-a, so why make a new name...

TiejunMS:
> hello, ARMv8-A is hard to type... other projects refers it as armv8-a, gcc's compile option is -march=armv8-a, so why make a new name... when pass the armv8-a from a environment variable, you have to do one more step: translate armv8-a to ARMv8-A eg. if [ THREADX_ARCH == ARMv8-A ] CFLAGS += -march=armv8-a end of if

Thanks for the suggestion! But ports_arch is a folder for internal development only. Users should use files from ports, ports_module and ports_smp.

ports/cortex_a35/gnu/example_build/tx/.cproject
```diff
@@ -39,7 +39,7 @@
                             							
                             <tool id="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.var.gcc-8.3.0-aarch64-elf.1755052156" name="GCC C Compiler 8.3.0 [aarch64-elf]" superClass="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.var.gcc-8.3.0-aarch64-elf">
                                 								
-                                <option id="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.option.cpu.281590127" name="CPU (-mcpu)" superClass="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.option.cpu" useByScannerDiscovery="false" value="cortex-a35" valueType="string"/>
+                                <option id="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.option.cpu.281590127" name="CPU (-mcpu)" superClass="com.arm.eclipse.cdt.managedbuild.ds5.gcc.baremetal.tool.c.compiler.option.cpu" useByScannerDiscovery="false" value="" valueType="string"/>
```
bo-ms, line 42: Could you explain why the value is set as null? thanks

bo-ms (approved):

ports/cortex_a35/gnu/example_build/tx/.cproject
TiejunMS: Use -march to support generic CPU types and simplifies the maintenance.

PR #278, opened 2023-06-21
SysTick initialization problem fixed in tx_initialize_low_level.s

Polarisru:
Problem with first system tick was detected, it needs much more time for the first tick as it's defined. The reason for this behavior is incorrect initialization of the SysTick timer in the port file for the Cortex-M0. It doesn't reset the SysTick Current Value Register despite the fact that its value is not initialized at startup (see https://developer.arm.com/documentation/dui0552/a/cortex-m3-peripherals/system-timer--systick). So if we have 0xFFFFFF (this register has 24-bit), it means we will get about 256*256*256 / 48000000 for the tact frequency of 48MHz to reach the zero, that makes 350ms delay at startup, which is what was observed on a real microcontroller.

## PR checklist

### DIFF
--- a/ports/cortex_m0/ac5/example_build/tx_initialize_low_level.s
+++ b/ports/cortex_m0/ac5/example_build/tx_initialize_low_level.s
@@ -176,6 +176,9 @@ _tx_initialize_low_level
 ;    /* Configure SysTick.  */
 ;
     LDR     r0, =0xE000E000                         ; Build address of NVIC registers
+    LDR     r1, =0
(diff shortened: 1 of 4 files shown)


Polarisru:
@microsoft-github-policy-service agree

billlamiework (changes requested):
Not a big issue, but it would be better to use the same comment style instead of the //

fdesbiens:
Merging this and bypassing the ECA since the changes are trivial.

PR #287, opened 2023-07-17
Added thumb mode support under GNU for module manager on Cortex-A7 pl…

yajunxiaMS:
Added thumb mode support under GNU for module manager on Cortex-A7 platform.


TiejunMS (approved):

ports_module/cortex_a7/gnu/inc/tx_port.h
```diff
@@ -273,17 +276,9 @@ typedef unsigned short                          USHORT;
 /* Determine if the ARM architecture has the CLZ instruction. This is available on
```
xiuwencai, line 276: Since the code is removed, how about removing this comment as well?

yajunxiaMS: accepted, resolved.

xiuwencai (approved):

PR #299, opened 2023-09-11
Fixed the issue of the ram cannot be read from ARM FVP debug tool in …

yajunxiaMS:
…cortex-A7 series port.

## PR checklist


yajunxiaMS:
All related code merged, close

PR #314, opened 2023-11-13
Update test script to generate JUnit format test report.

ting-ms:
- Update test script to generate JUnit format test report.
- Update the install script to upgrade CMake to the latest version.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -2,8 +2,11 @@
 #
 # Install necessary softwares for Ubuntu.
 
-apt-get update
-apt-get install -y \
+# Remove large folder to save space
+rm -rf /opt/hostedtoolcache
+
+sudo apt-get update
+sudo apt-get install -y \
     gcc-multilib \
     git \
     g++ \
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -2,8 +2,11 @@
 #
 # Install necessary softwares for Ubuntu.
 
-apt-get update
-apt-get install -y \
+# Remove large folder to save space
+rm -rf /opt/hostedtoolcache
```
bo-ms, line 6: Why do we delete this folder? is the space not enough?

scripts/install.sh
```diff
@@ -19,3 +22,5 @@ apt-get install -y \
 python3 -m pip install --upgrade pip
 pip3 install gcovr==4.1
 
+# Upgrade cmake to the latest version.
+pip install --upgrade cmake
```
TiejunMS, line 26: What is the required version?

ting-ms: The minimum required version is 3.21.

TiejunMS (approved):

bo-ms (approved):

Issue #334, opened 2023-12-19
Found a bug which causes tx_byte_allocate() not thread-safe when use -O3 optimization!!


yuhiping:
**The way we use byte pool**

We use byte pool as heap in our project,  the detail code in our project was copied from the following link which implemets CMSIS API
for threadx.
https://github.com/STMicroelectronics/stm32_mw_cmsis_rtos_tx/blob/main/cmsis_os2.c

In cmsis_os2.c, byte pool is created in **MemInit()** and after that, we can just use **MemAlloc()** and **MemFree()** to request and release memorry.
- **static uint8_t \*MemAlloc(uint32_t mem_size, uint8_t pool_type)**  was defined at line 158.
- **static osStatus_t MemInit(void)** was defined at line 247.
-  The *control block* or so called *handle* was defined at line 142 to 144 as **global varibles**:
**TX_BYTE_POOL HeapBytePool;**
**TX_BYTE_POOL StackBytePool;**
**TX_BLOCK_POOL BlockPool;**

**What happened?**
**MemAlloc()** and **MemFree()** just worked fine at most of the time. But occationally, **MemAlloc()** failed with code of **TX_NO_MEMORY** .  While, when we looked into this issue and after debugged a lot, we found that there was still enough memorry but the alloc algorithm not worked well.  

**Why this happened?**
After days of debugging, we found that the value of **tx_byte_pool_fragments** went wrong when this issue happen. 
To prove this, we defined a global varible of the same type, and do the same operation when **tx_byte_pool_fragments** increased or decreased. This is simple because **tx_byte_pool_fragments** only changes in the function *_tx_byte_pool_search()* which defined in threadx/threadx/common/src/tx_byte_pool_search.c.
It turned out that, when **MemAlloc()** failed with code of **TX_NO_MEMORY** , the value of **tx_byte_pool_fragments** in struct TX_BYTE_POOL was just different with the global varible we defined.

**The root cause**
After a deep dive into the source code of threadx, we found that **_tx_byte_pool_search()** can be interrupted after **TX_RESTORE** at line 258. This is a good design to let taskes with high priority alloc from the pool at first. But as mentioned above,  The *control block* or so called *handle* was defined as **global varibles**,  which may be changed by another high priority task.

**Further explanation**
1. After put a volatile before the **tx_byte_pool_fragments** varible, the issue was fixed and never came up.
2. When we changed the gcc compilation parameter from -O3 to -O1, this  issue was fixed and never came up.

[...]

PR #367, opened 2024-02-28
Fix the issue of byte pool alloc fail when compile with -O3 optimization

yuhiping:
https://github.com/eclipse-threadx/threadx/issues/334

## PR checklist


amgross (approved):

freyc:
It seems that the root-cause for this issue is a missing memory-clobber/compiler-barrier in the TX_DISABLE and TX_RESTORE macros (https://github.com/eclipse-threadx/threadx/issues/334#issuecomment-2068839866).
I would suggest fixing this issue in the cpu-specific ports instead. There might be other places in the code with possible issues because of this.

fdesbiens:
Superseded by #523.

PR #404, opened 2024-09-02
linux/gnu: add cmake for sample

yf13:
This adds CMake support for `linux/gnu/example_build` so that to have the `sample_threadx` target built after the `libthreadx.a` target, make it easier for people to try the sample on Linux.


fdesbiens:
Thank you for this, and apologies that it sat for so long.

The goal is right and it is still missing: the CMake build produces only the library, so trying ThreadX on Linux means using the Makefile instead. I have opened #622 to add it, derived from your two `example_build/CMakeLists.txt` files.

I am closing this one rather than rebasing it, and I would rather explain why than leave it looking like a rejection.

Four of the six files here have since arrived on `dev` by other routes: the top level's SMP selection, `common_smp/CMakeLists.txt`, `ports/linux/gnu/CMakeLists.txt` and `ports_smp/linux/gnu/CMakeLists.txt`. That is what the conflict is. Only the two sample targets were still missing, and those are what #622 carries.

Two details from the superseded part are worth passing back, since they may be useful to you elsewhere.

The SMP switch used `if(DEFINED THREADX_SMP)`, which is true whenever the variable exists at all, so `-DTHREADX_SMP=OFF` selected the SMP build:

| invocation | resulting suffix |
|---|---|
| `-DTHREADX_SMP=OFF` | `_smp` |
| no flag | *(empty)* |
| `-DTHREADX_SMP=ON` | `_smp` |

`dev` now uses `option()` with `if(THREADX_SMP)`, which treats `OFF` as off.

And the path scheme built `ports_smp/${THREADX_ARCH}/...` with no suffix on the architecture. That works for `linux` and `win64`, which predate the naming convention, but not for the ports named `<arch>_smp`. `dev` appends the suffix and falls back to the bare name for exactly those two.

One change was needed to your example files themselves: they read `common${THREADX_SMP}/inc`, which relied on `THREADX_SMP` being a string suffix. It is a boolean now, so that expands to `commonON`; #622 uses `${TX_COMMON_DIR}` instead, which the top level already resolves.

Both variants now build and run — the non-SMP demo prints its banner and counters, and the SMP one prints the SMP banner. Credit for the idea and the starting point is recorded in #622 and in its commit message. Thank you for the contribution, and for #403, which was merged.

PR #414, opened 2024-10-02
const: Add const keyword for all name pointer parameters

MaJerle:
This pull requests fixes the compiler warning when creating TX objects with literal strings since we are passing literal string to the function accepting `CHAR *`, that can, by design, modify input parameter and therefore cause undefined system behavior.

All name pointers become `const CHAR*` type.

There is no backward incompatibility when creating objects, but there is a potential backward in-compatibility when calling `tx_<modulename>_name_get` function, because parameter shall be changed to `const char* variable`.

Currently we can get the name as:
```c
CHAR* name;
tx_thread_name_get(..., &name);
```

New required function call:
```c
const CHAR* name;
tx_thread_name_get(..., &name);
```

I've not seen tests for getting the name in the threadx, I also clearly assume that people using `_get` api are the minority.
From safety and quality standpoint, this upgrade makes a lot of sense.

## PR checklist

- [x] Validated on real hardware C-M4, GCC

### DIFF
--- a/common/inc/tx_api.h
+++ b/common/inc/tx_api.h
@@ -428,7 +428,7 @@ typedef struct TX_TIMER_STRUCT
     ULONG               tx_timer_id;
 
     /* Define the timer's name.  */
-    CHAR                *tx_timer_name;
+    const CHAR          *tx_timer_name;
 
     /* Define the actual contents of the timer.  This is the block that
        is used in the actual timer expiration processing.  */
@@ -494,7 +494,7 @@ typedef struct TX_THREAD_STRUCT
        is typically defined to whitespace or a pointer type in tx_port.h.  */
     TX_THREAD_EXTENSION_0
 
-    CHAR                *tx_thread_name;                /* Pointer to thread's name     */
+    const CHAR          *tx_thread_name;                /* Pointer to thread's name     */
     UINT                tx_thread_priority;             /* Priority of thread (0-1023)  */
     UINT                tx_thread_state;                /* Thread's execution state     */
     UINT                tx_thread_delayed_suspend;      /* Delayed suspend flag         */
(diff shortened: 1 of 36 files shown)


fdesbiens:
@eclipse-threadx/iot-threadx-committers I need one of you to review this PR, please.

fdesbiens:
Thank you for this contribution, @MaJerle. I asked the group of committers to review it. I appreciate your patience. 

Before we can merge this PR, you need to sign the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/eca/). The purpose of the ECA is to provide a written record that you have agreed to provide your code and documentation contributions under the licenses used by the Eclipse ThreadX project. It also makes it clear that you are promising that what you are contributing to Eclipse is code you wrote, and you have the necessary rights to contribute it to our projects. And finally, it documents a commitment from you that your open source contributions will be permanently on the public record.

Signing the ECA requires an Eclipse Foundation account if you do not already have one. You can create one for free at [https://accounts.eclipse.org](https://accounts.eclipse.org). 

**Be sure to use the same email address when you register for the account that you intend to use on Git commit records.**

billlamiework:
In principle, I agree with this request, and we should have included this in the original APIs. However, now that the API has been public for 28 years, I would refrain from making this change. I don't like anything that alters the published API. That said, if we were to do this, we would also have to update the user guide and, I think, the name pointer storage in all the control blocks, e.g., TX_THREAD, TX_QUEUE, etc.

MaJerle:
> In principle, I agree with this request, and we should have included this in the original APIs. However, now that the API has been public for 28 years, I would refrain from making this change. I don't like anything that alters the published API. That said, if we were to do this, we would also have to update the user guide and, I think, the name pointer storage in all the control blocks, e.g., TX_THREAD, TX_QUEUE, etc.

I may sound nasty, but if it was not done correctly for `28` years, it doesn't mean it shouldn't be fixed now. The implementation can lead to bugs.

PR #429, opened 2025-01-09
Issue 426: Add resume and suspend hook extensions

greglink:
Existing ThreadX source has a very powerful hook for thread entry and exit. This hook allows tracking of processor utilization for each task (by observing when it enters and leaves the processor), as well as any other usage deemed appropriate by the developer.

No such equivalent hook(s) exist for the suspension and resumption of threads, making it difficult to observe how long from the time a thread becomes ready until it first enters the processor, and eventually is suspended as well.

This pull request adds such hooks. At the same time, it does not follow the format and structure of the entry and exit hooks, as those were written with the expectation that developers and users would not have direct access to the source code, and as such, the hooks are less performant than they could otherwise be. This pull request therefore uses a more direct approach to allow installation of callback extensions that should have no performance impact when not enabled.

### Work In Progress ###
This is not a mergeable commit, as it includes modifications to things that should not be committed upstream. Importantly, it shows two possible approaches to installing/including the hooks. Furthermore, I have not yet benchmarked any performance impact of a system with the hooks enabled vs a system without the hooks.

## PR checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,9 @@ target_sources(freertos-threadx
 )
 target_link_libraries(freertos-threadx PUBLIC threadx)
 
+
+set(TX_USER_FILE "${CMAKE_CURRENT_LIST_DIR}/custom_tx_user.h" CACHE PATH "Path to custom tx_user.h")
+
 # If the user provided an override, copy it to the custom directory
 if (NOT TX_USER_FILE)
     message(STATUS "Using default tx_user.h file")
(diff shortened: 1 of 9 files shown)


fdesbiens:
Dear @eclipse-threadx/iot-threadx-committers and @eclipse-threadx/iot-threadx-contributors, please review this potential contribution. The sooner you provide feedback to the author, the sooner we'll have a brand-new ThreadX feature. Thanks!

billlamiework:
Have you considered overriding the existing macro TX_THREAD_STATE_CHANGE with a call to whatever state change hook processing you need?   The state change macro looks like this:

#define TX_THREAD_STATE_CHANGE(a, b)

Where "a" is the TX_THREAD control block pointer and "b" is the new state for the thread. If this is sufficient, no code change is required.

PR #480, opened 2025-12-07
basic processing case shows no results with O2

hotislandn:
## PR checklist

When compiling with the arm-none-eabi-gcc 13.2.1 with "-O2" for v8-r, this test case just outputs "0" as the result.

### DIFF
--- a/utility/benchmarks/thread_metric/tm_basic_processing_test.c
+++ b/utility/benchmarks/thread_metric/tm_basic_processing_test.c
@@ -46,7 +46,7 @@
 
 /* Define the counters used in the demo application...  */
 
-unsigned long   tm_basic_processing_counter;
+volatile unsigned long   tm_basic_processing_counter;
 
 
 /* Test array.  We will just do a series of calculations on the 


hotislandn:
Well, we tried to do the compile only test with the latest code base of threadx with GCC and LLVM today.
The detail info is:
GCC: gcc version 14.2.1 20241119 (Arm GNU Toolchain 14.2.Rel1 (Build arm-14.52))  <--- new version here
Clang: version 21.1.0
ThreadX source code at commit: c4ad279b85fcf836c0b42d3e140fc32f0554e7b5

Compile cmd for GCC:
`threadx/utility/benchmarks/thread_metric$ arm-none-eabi-gcc -march=armv8-r -mfpu=fpv5-sp-d16 -mfloat-abi=hard -mlittle-endian -mthumb -fdata-sections -ffunction-sections -O2 -g3 -c tm_basic_processing_test.c`
and objdump shows that:
```
00000000 <tm_basic_processing_thread_0_entry>:
   0:   2300            movs    r3, #0
   2:   f240 0100       movw    r1, #0
   6:   461a            mov     r2, r3
   8:   f2c0 0100       movt    r1, #0
   c:   f841 2023       str.w   r2, [r1, r3, lsl #2]
  10:   3301            adds    r3, #1
  12:   f5b3 6f80       cmp.w   r3, #1024       @ 0x400
  16:   d1f9            bne.n   c <tm_basic_processing_thread_0_entry+0xc>
  18:   f240 0300       movw    r3, #0
  1c:   f2c0 0300       movt    r3, #0
  20:   f8d3 c000       ldr.w   ip, [r3]
  24:   2300            movs    r3, #0
  26:   f851 2023       ldr.w   r2, [r1, r3, lsl #2]
  2a:   f851 0023       ldr.w   r0, [r1, r3, lsl #2]
  2e:   4462            add     r2, ip
  30:   4042            eors    r2, r0
  32:   f841 2023       str.w   r2, [r1, r3, lsl #2]
  36:   3301            adds    r3, #1
  38:   f5b3 6f80       cmp.w   r3, #1024       @ 0x400
  3c:   d1f3            bne.n   26 <tm_basic_processing_thread_0_entry+0x26>
  3e:   f10c 0c01       add.w   ip, ip, #1
  42:   e7ef            b.n     24 <tm_basic_processing_thread_0_entry+0x24>
```
The instruction for "[tm_basic_processing_counter](https://github.com/eclipse-threadx/threadx/blob/master/utility/benchmarks/thread_metric/tm_basic_processing_test.c#L49)" is at address 3e.

Compile cmd for LLVM:
`clang --target=arm-none-eabi -march=armv8-r -mfpu=fpv5-sp-d16 -mfloat-abi=hard -mlittle-endian -mthumb -fdata-sections -ffunction-sections -O2 -g3 -c tm_basic_processing_test.c`
It generates:
```
00000000 <tm_basic_processing_thread_0_entry>:
   0:   f240 0000       movw    r0, #0
   4:   2100            movs    r1, #0
   6:   f2c0 0000       movt    r0, #0
   a:   2200            movs    r2, #0
   c:   f840 1022       str.w   r1, [r0, r2, lsl #2]
  10:   3201            adds    r2, #1
  12:   f5b2 6f80       cmp.w   r2, #1024       @ 0x400
  16:   d1f9            bne.n   c <tm_basic_processing_thread_0_entry+0xc>
  18:   f240 0c00       movw    ip, #0
  1c:   f2c0 0c00       movt    ip, #0
  20:   f8dc e000       ldr.w   lr, [ip]
  24:   2300            movs    r3, #0
  26:   f850 1023       ldr.w   r1, [r0, r3, lsl #2]
  2a:   f850 2023       ldr.w   r2, [r0, r3, lsl #2]
  2e:   4471            add     r1, lr
  30:   4051            eors    r1, r2
  32:   f840 1023       str.w   r1, [r0, r3, lsl #2]
  36:   3301            adds    r3, #1
  38:   f5b3 6f80       cmp.w   r3, #1024       @ 0x400
  3c:   d1f3            bne.n   26 <tm_basic_processing_thread_0_entry+0x26>
  3e:   f10e 0e01       add.w   lr, lr, #1
  42:   f8cc e000       str.w   lr, [ip]
  46:   e7ed            b.n     24 <tm_basic_processing_thread_0_entry+0x24>
```
The instructions at address 3e and 42 does the expected job, which is different from GCC.

BTW, O1 and O3 for GCC does not solve the problem here.

Hope those tedious logs helpful for this topic.

billlamiework:
Have you tried to add "volatile" to the "unsigned long   tm_basic_processing_counter;" definition to see if that keeps the variable in memory?

hotislandn:
> Have you tried to add "volatile" to the "unsigned long tm_basic_processing_counter;" definition to see if that keeps the variable in memory?

Hi Bill,

Yes, just as it showed in the patch. 
And with the "volatile", the code section is compiled to:
```
00000000 <tm_basic_processing_thread_0_entry>:
   0:   2300            movs    r3, #0
   2:   f240 0100       movw    r1, #0
   6:   461a            mov     r2, r3
   8:   b410            push    {r4}
   a:   f2c0 0100       movt    r1, #0
   e:   f841 2023       str.w   r2, [r1, r3, lsl #2]
  12:   3301            adds    r3, #1
  14:   f5b3 6f80       cmp.w   r3, #1024       @ 0x400
  18:   d1f9            bne.n   e <tm_basic_processing_thread_0_entry+0xe>
  1a:   f240 0c00       movw    ip, #0
  1e:   f2c0 0c00       movt    ip, #0
  22:   2300            movs    r3, #0
  24:   f851 2023       ldr.w   r2, [r1, r3, lsl #2]
  28:   f8dc 4000       ldr.w   r4, [ip]
  2c:   f851 0023       ldr.w   r0, [r1, r3, lsl #2]
  30:   4422            add     r2, r4
  32:   4042            eors    r2, r0
  34:   f841 2023       str.w   r2, [r1, r3, lsl #2]
  38:   3301            adds    r3, #1
  3a:   f5b3 6f80       cmp.w   r3, #1024       @ 0x400
  3e:   d1f1            bne.n   24 <tm_basic_processing_thread_0_entry+0x24>
  40:   f8dc 3000       ldr.w   r3, [ip]
  44:   3301            adds    r3, #1
  46:   f8cc 3000       str.w   r3, [ip]
  4a:   e7ea            b.n     22 <tm_basic_processing_thread_0_entry+0x22>
```
That variable is handled by lines 40/44/46(relative addresses) and is written to memory.

fdesbiens:
@hotislandn It looks like this conversation stalled. Since the variable has been written to memory when `volatile` is used I would update our documentation to recommend using this technique in similar contexts.

 I am also happy to update the following files to reflect that recommendation:

  | File | Counter Variable(s) to make `volatile` |
  | :--- | :--- |
  | `tm_basic_processing_test.c` | `tm_basic_processing_counter` |
  | `tm_cooperative_scheduling_test.c` |      `tm_cooperative_thread_0_counter`<br>`tm_cooperative_thread_1_counter`<br>`tm_cooperative_thread_2_counter`<br>`tm_cooperative_thread_3_counter`<br>`tm_cooperative_thread_4_counter`|
 | `tm_interrupt_preemption_processing_test.c` |      `tm_interrupt_preemption_thread_0_counter`<br>`tm_interrupt_preemption_thread_1_counter`<br>`tm_interrupt_preemption_handler_counter` |
  | `tm_interrupt_processing_test.c` | `tm_interrupt_thread_0_counter`<br>`tm_interrupt_handler_counter` |
  | `tm_memory_allocation_test.c` | `tm_memory_allocation_counter` |
  | `tm_message_processing_test.c` | `tm_message_processing_counter` |
  | `tm_preemptive_scheduling_test.c` |`tm_preemptive_thread_0_counter`<br>`tm_preemptive_thread_1_counter`<br>`tm_preemptive_thread_2_counter`<br>`tm_preemptive_thread_3_counter`<br>`tm_preemptive_thread_4_counter` |
 | `tm_synchronization_processing_test.c` | `tm_synchronization_processing_counter` |
 
 What do you think?

hotislandn:
Hi, @fdesbiens,

So far, we have only seen this issue in the case "tm_basic_processing_test.c". 
We did test the following cases:
tm_cooperative_scheduling_test
tm_memory_allocation_test
tm_message_processing_test
tm_preemptive_scheduling_test
tm_synchronization_processing_test
On the same HW, with the same toolchains/options. And those cases generated the expected results.

fdesbiens:
Thank you for this — your diagnosis was right, and the reproduction was precise enough that I could confirm it directly rather than guess at it.

I reproduced it with the same compiler you used, `arm-none-eabi-gcc 13.2.1 20231009`, at `-O2 -mcpu=cortex-r52`. The disassembly shows exactly what you inferred: the counter is loaded once before the loop, incremented in a register, and never stored, so the reporting thread reads a location that stays 0 for the life of the program.

```
  28:  ldr  ip, [r3]        counter loaded once
  50:  add  ip, ip, #1      increment stays in the register
  54:  b    2c              and around again
```

Two things I found while confirming it that are worth passing back to you.

It also explains a second symptom you may have seen: because the counter never changes, `tm_basic_processing_counter == last_counter` holds every period, so the test additionally prints "Basic processing thread died!". Same single cause.

And it is not specific to Armv8-R. The same shape appears for Cortex-M4 in Thumb state, and at `-O1`, `-O3` and `-Os` as well as `-O2`. Only `-O0` happens to work, because there the counter is spilled after every statement.

I have opened #616 with a slightly different fix, and I want to explain why rather than just replace your patch. `volatile` alone is correct for visibility, but leaving the counter in the loop expression means the compiler must re-read it on every one of the 1024 iterations — 3 loads per element instead of 2, 9 instructions instead of 8. The readme says this particular test is the baseline that every other result in the suite is scaled against, so changing its throughput rescales all cross-RTOS comparisons and breaks comparability with published figures. #616 keeps `volatile` and takes a snapshot into a local once per pass, which leaves the measured inner loop instruction-for-instruction identical to today while still publishing the counter.

I also checked the other seven tests, since they all share counters with the reporting thread without `volatile`. They are safe, and for a reason worth knowing: their loops call into the porting layer, and the compiler cannot prove the callee leaves the global alone, so it spills the counter around the call. Yours was the one test whose loop calls nothing at all — which is why it is the only one that fails.

I will close this PR in favour of #616, with the credit recorded there. Thank you for finding and reporting it.

hotislandn:
@fdesbiens,
Thanks for taking the time to investigate this so carefully!

I'm glad the reproduction helped confirm the issue. I also learned a lot from your analysis of the other cases and the way you validated the problem before deciding on the final solution.

Really appreciate the detailed explanation — it's a good lesson for me on being more thorough when diagnosing these kinds of issues. Thanks again!

PR #501, opened 2026-02-16
Remove dead code from riscv64/gnu port

akifejaz:
There was dead code in tx_thread_context_save.S the code after the return statement is unreachable. Removed that and also tested the demo application to make sure everything is fine.


akifejaz:
Hi @fdesbiens, this PR needs attention, it removes the dead code from risc-v64 port. The code after the return is unreachable and I've doubled checked its redundant there.

PR #508, opened 2026-03-04
Feature/rv64 rvv support

goodnorning:
## PR checklist

### DIFF
--- a/ports/risc-v64/gnu/example_build/qemu_virt/csr.h
+++ b/ports/risc-v64/gnu/example_build/qemu_virt/csr.h
@@ -21,6 +21,7 @@
 #define MSTATUS_MIE 		(1L << 3)    // machine-mode interrupt enable.
 #define MSTATUS_MPIE		(1L << 7)
 #define MSTATUS_FS		(1L << 13)
+#define MSTATUS_VS    (1L << 9)
 
 // Machine-mode Interrupt Enable
 #define MIE_MTIE		(1L << 7)  
--- a/ports/risc-v64/gnu/example_build/qemu_virt/demo_threadx.c
+++ b/ports/risc-v64/gnu/example_build/qemu_virt/demo_threadx.c
@@ -4,8 +4,13 @@
 
 #include "tx_api.h"
 #include "uart.h"
+#if defined(__riscv_vector)
+#define DEMO_STACK_SIZE         (1024 + 16448)      /* 16448 for RVV Extension */
+#define DEMO_BYTE_POOL_SIZE     (9180 + 148032)     /* 148032 for RVV Extension */
+#else
 #define DEMO_STACK_SIZE         1024
 #define DEMO_BYTE_POOL_SIZE     9180
+#endif
 #define DEMO_BLOCK_POOL_SIZE    100
 #define DEMO_QUEUE_SIZE         100
 
--- a/ports/risc-v64/gnu/example_build/qemu_virt/entry.s
+++ b/ports/risc-v64/gnu/example_build/qemu_virt/entry.s
@@ -41,7 +41,11 @@ _start:
 	li x30, 0
 	li x31, 0
 	la t0, _sysstack_start
+#ifdef __riscv_vector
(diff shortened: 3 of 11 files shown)


fdesbiens:
Thank you for this contribution, @goodnorning! We will review it and come back if there is any feedback.

We are currently under code freeze before the Q1 2026 release. I will only merge to dev after the release. This feature would ship in our Q2 2026 release in June.

akifejaz (commented):
Hi @goodnorning 

Thankyou so much for your contribution. 

I liked the implementation and it looks good from myside, Here is my feedback on some minor issues.

- vstart must be restored before vsetvl. vstart is the start index for vector operations.  Setting vstart AFTER vsetvl has no effect on the current vl configuration.
- saving  vxrm and vxsat separately which is functionally equivalent, but vcsr should also be explicitly handled for spec compliance.
- add `#ifdef __riscv_vector` guard in link.lds as its silently wasting 16 KB on every non-vector build.
-  the vxrm and vxsat seems swapped on restore, save stores vxrm at offset 3 and vxsat at offset 4, but every restore (csrw) writes offset 3 into vxsat and offset 4 into vxrm
-  also add `#ifdef __riscv_vector` guard in entry.s

fdesbiens:
@goodnorning Just to confirm you are OK with @akifejaz's feedback. I will merge to dev as soon as you make the suggested changes. Thank you once again for your contribution.

goodnorning:
> Hi @goodnorning
> 
> Thankyou so much for your contribution.
> 
> I liked the implementation and it looks good from myside, Here is my feedback on some minor issues.
> 
> * vstart must be restored before vsetvl. vstart is the start index for vector operations.  Setting vstart AFTER vsetvl has no effect on the current vl configuration.
> * saving  vxrm and vxsat separately which is functionally equivalent, but vcsr should also be explicitly handled for spec compliance.
> * add `#ifdef __riscv_vector` guard in link.lds as its silently wasting 16 KB on every non-vector build.
> * the vxrm and vxsat seems swapped on restore, save stores vxrm at offset 3 and vxsat at offset 4, but every restore (csrw) writes offset 3 into vxsat and offset 4 into vxrm
> * also add `#ifdef __riscv_vector` guard in entry.s

Hi @akifejaz 
Thank you for your careful review. Regarding the first point about vstart restore must be before vsetvl, I am a bit confused about this. Because after the vsetvl operation is performed, vstart will be reset to 0, so the previously saved value cannot be correctly restored.

fdesbiens:
Hi @goodnorning and @akifejaz 

Looking at the [current version of the vector spec](https://docs.riscv.org/reference/isa/unpriv/v-st-ext.html), I found the following statement:

> All vector instructions, including vset{i}vl{i}, reset the vstart CSR to zero.

Looking at the PR, here is how I understand the sequence.
```Assembly
  vsetvl zero, t2, t3   // Restore vl/vtype — this resets vstart to 0
  csrw vstart, t4        // Restore vstart — written AFTER vsetvl, so it sticks
```

The same logic applies to the vector register restore that follows — the `vsetvli` used to set up the element width for `vle8/vse8` operations would also zero `vstart`, so any `vstart` restore must come after all `vset*` instructions.

Could it be that this behaviour changed from pre-ratification versions of the extension, @akifejaz? 

In any case, if my analysis is correct, then @goodnorning is right.

akifejaz:
Thank you for correcting me. 

I think I meant saving in this order not restoring, sorry for typo. Please feel free to ignore this point. Thank you again.

goodnorning:
Hi @akifejaz 
The revisions have been incorporated per the review feedback. Please kindly review it once more at your convenience. Thank you. 
CC @fdesbiens

akifejaz:
> Hi @akifejaz The revisions have been incorporated per the review feedback. Please kindly review it once more at your convenience. Thank you. CC @fdesbiens

Looks good from myside, Thanks for the revision. 

Having Vector support in ThreadX is definitely an important milestone. :)

PR #522, opened 2026-04-14
Fixed MPIE being cleared leading to have interrupts being disable after an mret instruction

### LINKED ISSUE #517: Possible bug in RISC-V64 leading to no interrupts after MIE and MPIE bit have been cleared before an mret instruction
_**Describe the bug**_
In the RISCV-64 port, when returning from the trap handler, it seems like the thread being resumed has it's interrupts disabled. Both MPIE and MIE bits are cleared in the mstatus register. Leading to the resumed thread having no interrupts potentially leading to an infinite loop when polling a variable that changes inside an IRQ.

We are running ThreadX on the Polarfire SOC from Microchip. Using the latest RISCV-64 port, after an initial Hardware timer interrupt is issued and a call to both  **_tx_thread_context_save** **_tx_thread_context_restore** followed by a final **_mret_** instruction, we observe the _**mstatus**_ register having both the **MPIE** and **MIE** bits set to **0**, completely disabling any interrupt. Is this done by design ?

It happened in the following situation where at the startup we only have **1** thread running :

**_Init thread_**

1. HW timer is setup
2. An initial I2C request is being prepared 
3. HW Timer interrupt (call to tx context saving mechanisms)
4. Init thread resumes
5. I2C request is issued
6. HW Timer interrupt
7. Init thread resumes
8. Dead loop as we are waiting for an interrupt from the I2C driver.

Here is the relevant code where both **MPIE** and **MIE** are  being cleared 
https://github.com/eclipse-threadx/threadx/blob/3726d7906b4808bfec7855fc088e073199df9120/ports/risc-v64/gnu/src/tx_thread_context_restore.S#L146-L167

The bitwise not sets the bits at position **3(MIE)** and position **4(MPIE)** to **0**. Right after there is an **and** instruction with the original mstatus value stored in **t1**. Then finally, an **or** operation which just sets bits at position **11:12(MPP)**. A final **mret** instruction is issued. Since MPIE has been set to 0, **mret** will copy **MPIE** value back to **MIE** leading to interrupts being disabled.

There is additionally the register **t2** not being used but still being set. My guess is that it's part of some dead code now ? 

**To Reproduce**
Unfortunately, I can't share any screenshots or particular steps to reproduce the issue as I'm writing this from another laptop. I will inevitably share those once I get back to my work place if this issue catches your attention.

**Expected behavior**
After the call to **mret** instruction, the expectation would be to have the MIE bit not being cleared when returning to a thread which had it's interrupt enabled before it being context switched. The fact that the MPIE bit is cleared before returning from machine mode, makes is so that there is no other way to have the interrupts being enabled unless explicitly enabled back once inside a thread.

**Impact**
Infinite loop on polling a status variable from a driver which should change when an IRQ for this driver is called.

**Logs and console output**
Again, I will provide those as soon as possible

**Solution**

Maybe _or_ **~t2** with **t1** to at least save those MPIE and MIE bits.

`csrr    t1, mstatus`
`li      t2, ~0x1888`
`or      t1, t1, t2`

cpdpls:
Fixed #517

1. **MPIE** was being cleared alongside **MIE** before issuing an **mret** instruction. As a side effect, this led to have no interruptions enabled forever when returning from the very first context switching.
2. Removed some assembly leftovers from the previous ports when changing the **mstatus** register inside the _tx_thread_context_restore routine 

## PR checklist

- [ ] Validated on real hardware (Microchip Polarfire SOC -- riscv64-unknown-elf-gcc (xPack GNU RISC-V Embedded GCC (Microsemi SoftConsole build), 64-bit) 8.3.0


fdesbiens:
Thank you for the detailed bug report and the fix @cpdpls — the root cause analysis is correct. Clearing MPIE before mret is indeed the culprit: mret copies MPIE→MIE, so with MPIE=0 interrupts are permanently disabled after the first context switch.

However, the fix seems to introduce a new logic error in both modified hunks.

The problem
```Assembly
 li      t2, 0x1880        // Set MPP(0x1800) | MPIE(0x80)
 li      t3, ~0x8          // Preserve all bits except MIE   ← computed but never read
 
 or      t1, t1, t2        // MPP and MPIE are now set       ✅
 and     t1, t1, t2        // Clear MIE                      ❌ wrong register
```

The `and` uses `t2` (value `0x1880`) instead of `t3` (value `~0x8`). ANDing with `0x1880` zeros every mstatus bit except MPP and MPIE, corrupting fields like FS (bits 14:13), VS (bits 10:9), MPRV (bit 17), and others. The FP and vector #if blocks below happen to re-set FS and VS, partially hiding the problem in those build variants — but the base (no FP, no vector) build is left with a corrupted mstatus.

t3 is loaded but never used — the same dead-register issue the PR description notes about the original code.

The same logic appears in both the `_tx_thread_context_restore` and `_tx_thread_no_preempt_restore` hunks.

Suggested fix:

Since `~0x8 = -9` fits within RISC-V's 12-bit signed `andi` immediate range, `t3` can be dropped entirely:

```Assembly
 csrr    t1, mstatus
 li      t2, 0x1880        // MPP(0x1800) | MPIE(0x80)
 or      t1, t1, t2        // Set MPP and MPIE
 andi    t1, t1, ~0x8      // Clear only MIE (bit 3); preserve all other mstatus bits
 ```

Apply the same three-instruction sequence to both hunks.

What do you think?

cpdpls:
Hello @fdesbiens,

Indeed, I somehow reused T2 both inside the **AND** and the **OR** instruction. My intention was to use T3 with the **AND** instruction … I feel a bit ashamed that I let that one through though.

Thank you for reviewing though, I totally agree with the provided solution and will update the PR *immediately*.

Sorry again for that embarrassing mistake.

cpdpls:
PR has been updated with the new suggestions. Thanks again for the review

fdesbiens:
@cpdpls The embarrassing mistake is all ours. After all, you were just fixing our code... ;-) Thanks again.

PR #526, opened 2026-04-28
Add regression test BSP, build scripts, and CI for RV32/RV64

akifejaz:
Adds QEMU-virt BSP and regression test infrastructure for RISC-V 32/64-bit ports.
 
  - QEMU-virt board support package (CLINT timer, PLIC, UART, trap handler, linker script)
  - Build/test scripts (run.sh, build_tx_riscv.sh, test_tx_riscv.sh) running 5 configs × 2 architectures (950 tests)
  - CI integration via regression_test.yml using xPack toolchain + QEMU
  - Fix 64-bit portability issues in regression tests: block/byte pool sizes, byte pool search pointer offset, timing tolerance, -Wconversion in trace code
  
  
All regression tests passes for RV32/RV64, please see the pipeline status [here ](https://github.com/akifejaz/threadx/actions/runs/25038795732/job/73336659336)

### DIFF
--- a/test/tx/cmake/riscv/bsp/plic.c
+++ b/test/tx/cmake/riscv/bsp/plic.c
@@ -1,0 +1,87 @@
+/***************************************************************************
+ * Copyright (c) 2026 10xEngineers
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License which is available at
+ * https://opensource.org/licenses/MIT.
+ *
+ * SPDX-License-Identifier: MIT
+ **************************************************************************/
+
+#include "plic.h"
+#include <stddef.h>
+
+irq_callback callbacks[MAX_CALLBACK_NUM];
+
+void plic_irq_enable(int irqno)
+{
+    uintptr_t hart = riscv_get_core();
+    volatile uint32_t *reg = (volatile uint32_t *)(PLIC_MENABLE(hart) + (irqno / 32) * 4);
+    *reg = *reg | (1U << (irqno % 32));
+}
+
+void plic_irq_disable(int irqno)
+{
+    uintptr_t hart = riscv_get_core();
+    volatile uint32_t *reg = (volatile uint32_t *)(PLIC_MENABLE(hart) + (irqno / 32) * 4);
+    *reg = *reg & ~(1U << (irqno % 32));
+}
+
+void plic_prio_set(int irqno, int prio)
+{
+    PLIC_SET_PRIO(irqno, prio);
+}
+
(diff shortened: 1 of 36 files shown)


test/tx/cmake/riscv/bsp/plic.c
```diff
@@ -0,0 +1,80 @@
...
+void plic_complete(int irqno)
+{
+    uintptr_t hart = riscv_get_core();
+    *(uint32_t *)(PLIC_MCOMPLETE(hart)) = (uint32_t)irqno;
+}
+
+int plic_irq_intr(void)
+{
+    int ret = -1;
+    int irqno = plic_claim();
```
fdesbiens, line 77: `plic_claim()` can return any value 0–1023, but `callbacks[]` only has `MAX_CALLBACK_NUM=128` entries. There is no bounds check before `callbacks[irqno]` is accessed.

```C
 // Fix: add bounds check at top of plic_irq_intr()
 int irqno = plic_claim();
 if (irqno <= 0 || irqno >= MAX_CALLBACK_NUM) {
     plic_complete(irqno);
     return 0;  // spurious or out-of-range, not an error
 }
```

Note: checking `irqno == 0` also handles the PLIC spurious-interrupt case (irqno=0 means "no interrupt").

test/tx/cmake/riscv/bsp/plic.c
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ **************************************************************************/
+
+#include "plic.h"
+#include <stddef.h>
+
+irq_callback callbacks[MAX_CALLBACK_NUM];
+
+void plic_irq_enable(int irqno)
```
fdesbiens, line 16: `(1 << irqno)` is UB when `irqno >= 32 (int overflow)`. The PLIC enable registers are also word-arrays, so only IRQs 0–31 are correctly addressable. 

Fix: use `(1U << (irqno % 32))` with the correct word offset, or add a guard that asserts/returns if `irqno >= 32`.

fdesbiens (changes requested):

fdesbiens:
@akifejaz I reviewed the PR and everything looks good. Please add the two small comments above on the PLIC, and we should be good to go!

akifejaz:
> @akifejaz I reviewed the PR and everything looks good. Please add the two small comments above on the PLIC, and we should be good to go!

Thank you for the review, I've updated the PR as per your comments. PR is ready from myside.

PR #529, opened 2026-05-08
Added win64 ports of ThreadX and ThreadX SMP

fdesbiens:


fdesbiens:
@billlamiework and @cypherbridge

This PR add brand-new Win64 ports of ThreadX and ThreadX SMP. There are now PowerShell scripts to build the code and run the regression tests for them.

At this point, all tests pass reliably for both versions. Would you mind reviewing the code, please? 

If you want to run them, you will need Visual Studio Build Tools 2022 (I used 19.42.34436.0) or any Visual Studio installation bundling them. The code targets Windows 11 only, as other versions are obsolete at this point. I will probably update the win32 port to use the same compiler later.

PR #534, opened 2026-05-19
Reverted ULONG to be 4 bytes long, fixing missaligment issue in the current port

### LINKED ISSUE #520: Regression tests cannot be generalized and be used with other architectures
The current tests were written with some assumptions that do not work out for other architectures.

In getting the tests working on a virtual RISC-V platform(32 and 64 bit), I ran into the following issues:

- counters are not marked volatile, at higher optimization levels some tests deadlock
- some memory tests do not provide pools large enough to accommodate the required amount of blocks when working with a 64-bit system. Due to the overhead being larger the pools end up providing too few blocks, thus failing the tests
- some pointers are only [4 byte aligned](https://github.com/eclipse-threadx/threadx/blob/3726d7906b4808bfec7855fc088e073199df9120/test/tx/regression/threadx_trace_basic_test.c#L347), again not working on a 64-bit system which requires 8 byte alignment
- in some instances, there is "resonance" and the test does not progress. The test expects an interrupt during a small window, but I encountered cases where this window is always missed. This is exacerbated by the fact that on a virtual platforms  instructions are always taking the same amount of time. In one instance there already is code to [include a perturbation](https://github.com/eclipse-threadx/threadx/blob/3726d7906b4808bfec7855fc088e073199df9120/test/tx/regression/threadx_event_flag_isr_set_clear_test.c#L75), but in my setup this was not large enough to make the test complete in a sensible time

While some of these points can be addressed pretty straight-forward, I would like to start a discussion, especially on the resonance issue.

cpdpls:
This PR should fix #520. After the recent internal meeting, here is the solution we came up with. The riscv64 port now follows the same type size as other 64-bit architectures in tx_port.h. This also fixes some misalignment issues reported in #520.

This also makes other ThreadX components such as NetX working again (ULONG was assumed to be 4 bytes and not 8 bytes as the current official port).

## PR checklist

- [x] Validated on real hardware (Microchip Polarfire SOC -- riscv64-unknown-elf-gcc (xPack GNU RISC-V Embedded GCC (Microsemi SoftConsole build), 64-bit) 8.3.0 and (Qemu Risc-V64 emulator 11.0 )


fdesbiens:
@akifejaz Please review.

akifejaz:
> @akifejaz Please review.

Sure, I'll review shortly.


Hi @cpdpls, there is PR which enables regression for risc-v ports. It has bunch of tests that runs on both RV32 and RV64. It would be nice if you can try running those. 

here is how you can run tests..
```
cd ~/threadx
./scripts/build_tx_riscv.sh 
./scripts/test_tx_riscv.sh
```

you would need to merge the upstream/dev branch into yours before running tests.

cpdpls:
Hi @akifejaz 

That was literally the question I was about to ask, so your comment comes in super handy ! I'll surely run the test. Thank you for your input !

fdesbiens:
Thank you @akifejaz and @cpdpls! 

I reviewed the changes, and they are in line with what the Arm ports provide. Well done! I will merge this for now.

@cpdpls Please submit fixes if your test run(s) find anything.

cpdpls:
Hi @fdesbiens,

Sorry I’ve been a tiny bit busy for this week/week-end. I’ll run a test by this week for sure and confirm back to you if the regression tests are passing. Sorry for the delay ! 

Have a nice starting week

cpdpls:
@fdesbiens FYI, I'm currently running into an issue where the command to build threadx as a standalone library does not work anymore because **THREADX_ARCH** is not defined. I guess this is an aftermath of removing the **project** directive at the top of the CMakeLists.txt in a recent commit.

fdesbiens:
@cpdpls I merged #540 to address this issue. In the future, please create an issue to track such problems. Commenting on a closed PR is not the best way to get my attention. ;-)

cpdpls:
My mistake, I will think about it next time ! :)

PR #513, opened 2026-03-13
arch/risc-v32: Implement lazy FPU, GP relaxation, and QEMU automation for GNU port

Winstonllllai:
This PR focuses on technical debt cleanup and functional enhancements for the **ThreadX RISC-V 32-bit (RV32) port**.
* **Symbolic Stack Offsets**:
    * Replaced hardcoded magic numbers in `.S` files with symbolic macros in `tx_port.h`.
    * Decoupled stack layout for easier future updates (e.g., Vector or Lazy FPU extensions).
* **Scheduler & Preemption Fixes**:
    * Refactored `_tx_thread_context_restore` to resolve potential race conditions in high-priority task preemption.
    * Integrated the `wfi` (Wait For Interrupt) instruction to optimize CPU idle cycles and power efficiency.
* **Automated Testing Framework**:
    * Introduced a Python-based E2E test harness using **QEMU + GDB**.
    * Validates multitasking, timer interrupts, FPU context preservation, and priority consistency.
* **Linker & Performance Optimization**:
    * Defined `__global_pointer$` in `link.lds` to enable **GP Relaxation**.
    * Reduced code size and improved global variable access efficiency.

## How to Run Tests
To verify the changes using the automated framework, please run the following commands:
```bash
cd build_qemu
make check-functional-riscv32
```
This is my first PR in Threadx. Please let me know if there’s anything I should improve or if I missed anything.

PR #549, opened 2026-06-09
arch/risc-v64: Add lazy FPU stacking and QEMU functional tests for GNU port

Winstonllllai:
This PR adds functional enhancements for the ThreadX RISC-V 64-bit (RV64) GNU port, mirroring the improvements delivered for RV32 in #513.

**Lazy FPU Stacking:**
- Updated `_tx_thread_context_save` and `_tx_thread_context_restore` to save `mstatus`/`sstatus` and skip FP register save/restore when the FS field (bits 14:13) is Off.
- Applied to nested save, non-nested save, nested restore, no-preempt restore, and preempt restore paths.
- Supports `TX_RISCV_SMODE` (reads `sstatus` vs `mstatus` as appropriate).
- Reduces context switch overhead for threads that do not use floating point.

**Automated Testing Framework:**
- Introduced a Python-based E2E test harness using QEMU + GDB (`azrtos_test_tx_gnu_riscv64_qemu.py`).
- Added `qemu_virt/CMakeLists.txt` and registered the `check-functional-riscv64` target.
- Validates multitasking, timer interrupts, FPU context preservation, and priority-based preemption.
- Extended `demo_threadx.c` with `fpu_test_val` and adjusted thread timing for GDB-driven verification.
**QEMU Boot & Link Fixes:**
- Placed `_start` in a `.text.boot` section and added `KEEP(*(.text.boot))` to `link.lds` so the entry point is pinned at `0x80000000` (QEMU virt reset PC).
- Linked `kernel.elf` with `--whole-archive` to ensure all ThreadX symbols resolve on RV64.
## How to Run Tests
```bash
cmake -B build_qemu -DCMAKE_TOOLCHAIN_FILE=cmake/riscv64_gnu.cmake
cd build_qemu
make check-functional-riscv64
```
Tested on QEMU virt machine (rv64gc).
Please let me know if there is anything I should improve or if I missed anything.


fdesbiens:
Hi @Winstonllllai.

Thank you for this contribution. @akifejaz and I will review it.

Winstonllllai:
Hi @fdesbiens and @akifejaz ,
Thank you both for taking the time to review this contribution. I appreciate it and look forward to your feedback.

akifejaz:
Hi @Winstonllllai please see comments below

akifejaz (changes requested):
Hi @Winstonllllai I see multiple issue at first glance some of those I already pointed in your previous PR : https://github.com/eclipse-threadx/threadx/pull/513

and 
- i see no significance for test/ dir & *.py script you can keep it for local testing. -- please also remove from RV32 port (PR : https://github.com/eclipse-threadx/threadx/pull/513)
- the "TX_RISCV_SMODE" is obviously from some other PR, seems like merging issue, you should not be adding these changes
- There are multiple new empty asm symbols in there, I'm not sure why you've added those if you chose to keep them empty. 
- also please change the comments carefully, I see you updated the comments wrongly, the earlier versions were fine.
- the file name is `entry.S` but your cmake calls it `entry.s' (small s), it should give build fails, please double check changes when submit PR

Also AI generated code should be marked properly. @fdesbiens when you get time, please also share the AI coding guidelines (if we have any) to him.

Winstonllllai:
Hi @akifejaz . Thanks for the review — I've addressed the points below:

1. **test/ & .py** — Removed from this PR; kept locally only. Will do the same for RV32 (#513).
2. **TX_RISCV_SMODE** — Removed the unintended sstatus branches; lazy FPU now uses mstatus only, aligned with RV32.
3. **Empty asm labels** — These are branch targets for the lazy FPU skip path (FS == Off), not standalone empty symbols.
4. **Comments** — Reverted incorrect changes; now reference mstatus only.
5. **entry.s** — QEMU CMake integration removed, so that issue is gone.
6. **AI-generated code** — Only the QEMU test script was AI-assisted; it has been removed from this PR.

The PR is now scoped to lazy FPU context save/restore only.

Winstonllllai:
For the RV32 cleanup requested in review, I opened a separate follow-up PR: #552. It removes the local-only QEMU test script and related CMake/test-only changes from the RV32 port.

fdesbiens:
Thank you for the cleanup — the scope now matches the title, and I can see every one of @akifejaz's points addressed:

- the `test/` directory and the `.py` script are gone; this PR now *removes* the RV64 one from `dev`
- `TX_RISCV_SMODE` is out of the diff. His read was right, and the feature has since reached `dev` by its own route
- the five remaining labels are all branch targets for the skip paths, not empty symbols
- the comment edits are confined to describing the new behaviour
- the `entry.S` / `entry.s` case mismatch went with the CMake file

I checked the mechanism rather than reading it. The `FS` extraction is correct — `srli 13; andi 0x3; beqz` isolates `mstatus[14:13]` and skips when Off — and slot 29 is in bounds for both ABIs, since the trap entry allocates 65 slots with floating point and 32 without. The whole RV64 port assembles clean with `riscv64-unknown-elf-gcc 15.2.0`, 8 of 8 files, hard float and soft float alike.

## One thing to fix before this can go in

Slot 29 is read uninitialised on the first restore of every newly created thread.

The frame map at the top of `tx_thread_stack_build.S` documents it as reserved:

```
                    x1      28      Initial ra
                    --      29      reserved
                    mepc    30      Initial mepc
```

and the builder skips it accordingly — it writes slot 28 and slot 30 and nothing in between, because until now nothing read slot 29.

This PR reads it. Since `mstatus` is never written back to the CSR, the slot is purely a "was FP live?" hint, so the first `_tx_thread_context_restore` for a new thread tests whatever was already on that stack. If the leftover bits happen to give `FS == 0`, the floating-point registers are not restored, and the thread starts with the FP state of whatever ran before it rather than the zeros that `tx_thread_stack_build.S` writes into slots 31 to 62. The outcome depends on stack contents, so it will look intermittent.

One line alongside the existing initialisers fixes it:

```asm
    li      t1, 0x2000                                  // mstatus.FS = Initial
    sd      t1, 29*8(t0)                                // Initial mstatus
```

Any non-zero `FS` works; `Initial` is the honest description of a thread that has not run yet. Worth updating the frame map comment from `reserved` to `mstatus` in the same change, so the next person reading it knows the slot is now live.

Please also add a test that a thread created *after* another thread has used floating point starts with clean FP state — that is the case this bug breaks, and it is the one a reviewer cannot see by reading.

fdesbiens:
@akifejaz — good catch, and yes, we do have guidelines. The authority is the Eclipse Foundation rather than anything we invent per project:

- [Generative Artificial Intelligence Usage Guidelines for Eclipse Committers](https://www.eclipse.org/projects/guidelines/genai/)
- the [Eclipse Project Handbook](https://www.eclipse.org/projects/handbook/), which has a "Contributing Content Generated by an AI" section
- the [Committer Due Diligence Guidelines](https://www.eclipse.org/legal/committer-guidelines/), which apply to AI-assisted contributions exactly as they do to any other

The short version for contributors: you may use these tools, the contribution still has to pass the same IP due diligence as anything else, and the committer merging it remains responsible for its correctness and provenance. Disclosure is expected rather than optional.

Concretely, here is what that looks like in code already on `dev`.

**A new file** carries the disclosure in its header. From `scripts/check_clang.sh`:

```sh
#!/bin/bash
##############################################################################
# Copyright (c) 2026 Eclipse ThreadX contributors
#
# This program and the accompanying materials are made available under the
# terms of the MIT License which is available at
# https://opensource.org/licenses/MIT.
#
# AI Disclosure: This file was largely AI-generated by Claude Code (Opus 5).
# The AI-generated portions may be considered public domain (CC0-1.0)
# and not subject to the project's licence. The human contributor has
# reviewed and verified that the code is correct.
#
# SPDX-License-Identifier: MIT and CC0-1.0
##############################################################################
```

Note the `SPDX-License-Identifier` becomes `MIT and CC0-1.0`, not plain `MIT`.

**An edit to an existing file** keeps that file's licence header and adds one line just below it, plus the Eclipse copyright if it is not already there. From `ports_module/cortex_m7/gnu/module_manager/src/tx_thread_schedule.S`:

```c
/***************************************************************************
 * Copyright (c) 2024 Microsoft Corporation
 * Copyright (c) 2026-present Eclipse ThreadX contributors
 *
 * This program and the accompanying materials are made available under the
 * terms of the MIT License which is available at
 * https://opensource.org/licenses/MIT.
 *
 * SPDX-License-Identifier: MIT
 **************************************************************************/
// Some portions generated by Codex (GPT-5).
```

Name the actual tool and version rather than a generic label — the tree currently has files naming Claude Code, Codex and Gemini, for example `ports_module/cortex_m7/gnu/module_manager/src/tx_thread_system_return.S`, which reads `// Some portions generated by Gemini (Gemini 2.0 Flash).`

@Winstonllllai — nothing to redo retroactively here; if any of this PR was AI-assisted, please add the appropriate marking to the two files you touched.

You are right that this is hard to find today: `CONTRIBUTING.md` says nothing about it. That file is due a broader refresh across the Eclipse ThreadX repositories, and these guidelines will be part of it. The updated version ships with our next release at the end of September.

Winstonllllai:
Hi @fdesbiens,
 
Thanks for the thorough review — I've addressed all the points below.
 
**Slot 29 initialization**
Added the missing initializer in `tx_thread_stack_build.S`:
```
li      t1, 0x2000                                  // mstatus.FS = Initial
sd      t1, 29*8(t0)                                // Initial mstatus
```
Also updated the frame map comment for slot 29 from `reserved` to `mstatus`, so it's documented as live going forward.
 
**Regression test**
Added `threadx_riscv_new_thread_fpu_state_test.c` under `test/tx/cmake/riscv/regression/`, registered in that directory's `CMakeLists.txt` gated on `THREADX_ARCH STREQUAL "risc-v64"` (RV32's stack builder doesn't write slot 29 yet, so the test doesn't apply there). It runs alongside the existing regression suite under ctest/QEMU like the other tests.
 
Test logic:
- `thread_0` dirties f0–f31 via inline asm to simulate a thread that has used the FPU.
- `thread_1` is created with `TX_DONT_START`, and before it's scheduled, the test inspects its freshly-built stack frame directly:
  - **ERROR 3** — slot 29 still holds the stack fill pattern (builder never wrote it).
  - **ERROR 4** — the FS field reads Off (frame incorrectly claims FP restore isn't needed).
- `thread_1` is then resumed and preempts `thread_0`; on entry it saves all 32 FP registers and checks they're all zero.
  - **ERROR 6** — any non-zero register means `thread_1` inherited `thread_0`'s FP state.
Note: the register-content check (ERROR 6) is compiled in only under hard-float ABIs (`__riscv_float_abi_double`/`single`); all four of our build configurations are hard-float, so ERROR 6 is exercised in every one of them. I ran the test across all four configurations and it passes in each with the fix in place, and fails as expected (via ERROR 6) if the two initializer lines are removed.
 
**AI disclosure**
The regression test file (`threadx_riscv_new_thread_fpu_state_test.c`) was AI-assisted (Claude Opus 5); I've added the appropriate disclosure header to that file per the guidelines you linked. No other files in this PR were AI-assisted.
 
Let me know if anything else needs adjustment before this can go in.

fdesbiens:
@Winstonllllai — picking this back up. Your lazy FPU work is good and it is going in; I have merged `dev` and pushed one commit to finish it. Your five commits are untouched.

### Why this needed a merge rather than a rebase

A lot happened on `dev` while this was open. #698 rewrote much of the RISC-V64 assembly you touch (frame sizes now come from a published contract in `tx_port.h`, the FP conditionals moved from ABI macros to `__riscv_flen`), and #691 did the same for RISC-V32. The branch was 121 commits behind.

Your `tx_thread_context_save.S` and `tx_thread_stack_build.S` changes merged **cleanly** with all of that, which is a good sign for how they were written.

### One thing I reverted

This branch deletes `ports/risc-v64/gnu/example_build/qemu_virt/test/threadx_test_tx_gnu_riscv64_qemu.py`, following @akifejaz's June request to keep the runners local.

That request has since been overtaken: he rebuilt those runners himself in #691 and #698, and they are now the functional verification for both RISC-V ports — MEPC integrity, time-slice, system clock and preemption, with a soft-float skip path. Deleting it now would remove test infrastructure that landed this week. I kept `dev`'s version. Nothing you did wrong; the ground moved.

### The commit I added

The merge left the feature half-applied, and half-applied is worse than not at all here. Your save side skips the FP stores when `mstatus.FS` is Off and records that judgement in frame slot 29 — but the three places that *consume* an interrupt frame still reloaded the FP registers unconditionally, out of slots the save had deliberately left alone. A thread that never touched the FP unit would have had stale frame contents loaded into its registers and `FS` driven to Dirty on the way out.

The guard is now at all three: both paths in `_tx_thread_context_restore`, and `_tx_thread_schedule_loop`. Same shape the risc-v32 port already uses.

I left the solicited path alone on purpose. `_tx_thread_system_return` saves the callee-saved FP registers unconditionally, so restoring them unconditionally is self-consistent — making that pair lazy too is a separate change, and risc-v32 is the model when someone wants it.

### Verified

| Suite | Result |
| --- | --- |
| RISC-V64, 5 configurations | **96 of 96**, nothing unlinkable |
| RISC-V32, 5 configurations | **95 of 95**, unchanged |
| `check-functional-riscv64` | every check passes |

The ninety-sixth test is yours. I checked it is load bearing rather than passing regardless: seeding stack build with `FS = Off` instead of `Initial` makes it fail, and restoring the seed makes it pass. It guards exactly the behaviour this branch is about.

That is worth saying plainly — a regression test that fails when the thing it tests is broken is the part most easily skipped, and you wrote one.

Issue #557, opened 2026-06-23
Implement mprotect()-based module isolation for the Linux/GNU simulation port (and Windows equivalent)


fdesbiens:
## Summary

The Linux/GNU simulation target has no module port. Adding one as a **test scaffold only** (no memory isolation) would allow the module manager sources to compile and be unit-tested on Linux but would not implement:

- Memory isolation (no `mprotect()`/MPU enforcement)
- The privileged dispatcher gate (no `SVC`-equivalent)
- Module binary loading from filesystem images
- Bounds checking on module-to-kernel pointer parameters

The goal of this issue is to go further and implement real `mprotect()`-based isolation in the Linux simulation port, making it useful for module boundary validation, developer tooling, and fuzzing.

## Proposed implementation: `mprotect()`-based isolation

Use `mprotect()` as a software analog to the ARM MPU:

| Embedded (ARM Cortex-M) | Linux simulation |
|---|---|
| 8–16 MPU region registers | Two `mprotect()` region descriptors |
| `SVC` instruction → SVC handler | C dispatcher gate with `mprotect()` permission swap |
| Hardware MPU reload on context switch | `mprotect()` syscall (~1 µs per switch) |
| MEMFAULT exception | `SIGSEGV` → `sigaction()` handler |

### New directory structure

```
ports_module/linux/gnu/
├── inc/
│   └── txm_module_port.h                        # port header
└── module_manager/src/
    ├── txm_module_manager_mm_setup.c            # mmap regions, mprotect enforcement
    ├── txm_module_manager_user_mode_entry.c     # replaces SVC gate
    └── txm_module_manager_memory_fault.c        # SIGSEGV handler → module fault callback
```

`TXM_MODULE_MANAGER_PORT_EXTENSION` would add two `mprotect` region descriptors to `TXM_MODULE_INSTANCE` (code region + data region).

### Dispatcher gate sketch

```c
ALIGN_TYPE _txm_module_manager_user_mode_entry(ULONG request, ...)
{
    ALIGN_TYPE result;
    mprotect(module_code, code_size, PROT_NONE);
    mprotect(module_data, data_size, PROT_NONE);
    result = _txm_module_manager_kernel_dispatch(request, ...);
    mprotect(module_code, code_size, PROT_READ | PROT_EXEC);
    mprotect(module_data, data_size, PROT_READ | PROT_WRITE);
    return result;
}
```

### Fault detection

```c
[...]

PR #558, opened 2026-06-23
Fixed module object deallocation of live kernel objects

fdesbiens:
## Summary

`_txm_module_manager_object_deallocate()` could free pool memory that still backs a live kernel object (TX_TIMER, TX_THREAD, TX_SEMAPHORE, TX_MUTEX, TX_QUEUE, TX_EVENT_FLAGS_GROUP, TX_BLOCK_POOL, TX_BYTE_POOL) without invoking the appropriate cleanup service first. This left stale kernel-list entries and timer-wheel entries pointing into freed/reused memory.

## Root cause

The function validated only that the memory belonged to the module instance's object pool before calling `_txe_byte_release()`. It did not check whether the memory still backed a live kernel object. By contrast, the stop path in `txm_module_manager_stop.c` already correctly iterates kernel created-lists and calls `tx_*_delete()` per type before releasing pool memory.

## Fix

Every TX kernel object struct stores its type ID as the **first ULONG** (e.g., `tx_timer_id`, `tx_semaphore_id`). `tx_*_delete()` clears this field to `TX_CLEAR_ID` (0) on deletion. The fix reads that field and dispatches to the appropriate `tx_*_delete()` (plus `_tx_thread_terminate()` for threads) before releasing pool memory. Already-deleted objects have ID 0 and are passed through safely.

**TX_THREAD** follows the same pattern as the stop path: `_tx_thread_terminate()` then `_tx_thread_delete()`.

## Backward compatibility

Default behavior is unchanged: the cleanup is silent and `TX_SUCCESS` is returned. Define `TXM_MODULE_OBJECT_DEALLOC_STRICT` (in `txm_module_user.h`) to receive `TX_DELETE_ERROR` whenever automatic cleanup is performed, signalling API misuse to the caller.

## Also included

- **Linux/GNU module simulation port** (`ports_module/linux/gnu/inc/txm_module_port.h`): minimal shim to allow module manager sources to compile and unit-test on Linux. Carries a prominent warning that no memory isolation is enforced. A real `mprotect()`-based isolation implementation is tracked in #557.
- **Regression tests**: 10 test cases covering all 8 object types, already-deleted objects, uninitialized memory, and both default and strict-mode variants. Verified passing on Linux.
- **Linux port extension**: `TX_THREAD_EXTENSION_2` gains `tx_thread_module_instance_ptr` (was empty) so the module manager compiles on Linux.

## Files changed

| File | Change |
|---|---|
| `common_modules/module_manager/src/txm_module_manager_object_deallocate.c` | Core fix |
| `common_modules/inc/txm_module_user_sample.h` | Documents `TXM_MODULE_OBJECT_DEALLOC_STRICT` |
| `ports/linux/gnu/inc/tx_port.h` | Adds `tx_thread_module_instance_ptr` to `TX_THREAD_EXTENSION_2` |
| `ports_module/linux/gnu/inc/txm_module_port.h` | New: Linux module simulation port scaffold |
| `test/tx/regression/threadx_module_object_dealloc_live_test.c` | New: regression tests |
| `test/tx/cmake/regression/CMakeLists.txt` | New test targets |


fdesbiens:
Closing. On further analysis, txm_module_object_deallocate is a deprecated API. The correct guidance is for module authors to call the appropriate tx_*_delete service, which the Module Manager dispatch layer already pairs with automatic pool deallocation. Hardening the deprecated function against misuse does not address the root cause. The associated security advisory has been closed as not a security vulnerability.

PR #560, opened 2026-06-24
Add SysTick counter reset in Cortex-M ports

hefanbo:
The SysTick counter value (NVIC offset 0x18) is indeterminate after reset. Without clearing it prior to enabling the timer, the first tick interval becomes unpredictable.

## PR checklist

- [x ] Validated on real hardware


hefanbo:
Tested on Cortex-M4 with gcc

hefanbo:
Commit message needs to be corrected

PR #561, opened 2026-06-24
Add SysTick counter reset in Cortex-M ports

hefanbo:
The SysTick counter value (SYST_CVR @0xE000E018) is indeterminate after reset. Without clearing it prior to enabling the counter, the first tick interval becomes unpredictable.

Tested on Cortex-M4 with gcc.

## PR checklist


fdesbiens:
Hi @hefanbo — thank you for this fix! The SysTick CVR indeterminacy at reset is a real hardware issue, and your solution (disable → clear → configure → enable) is exactly what the ARM TRM recommends. The fix is well-structured and consistent across all the affected ports.
 
 I took the liberty of pushing one small follow-up commit to your branch: it adds the missing inline comment on the `LDR r1, =0` line (to match the style of the other ports) and a required attribution header line per our contribution guidelines.
 
 Great contribution — much appreciated!

I will merge to dev now, and this will ship with our Q3 2026 release in September.

PR #639, opened 2026-08-18
Added a ThreadX module manager port for the Cortex-R52

fdesbiens:
A ThreadX module manager port for the Cortex-R52 under `ports_module/cortex_r52/gnu/`:
PMSAv8-R Stage-1 isolation of User-mode modules from the kernel and from each other, with an
example manager and module for the Armv8-R AEM FVP and the NXP S32Z280-594EVB. No GNU module
port existed for any R-profile core, so there was no close template.

## Design

**Seeded from `cortex_m33`, not `cortex_r4`.** PMSAv8-R uses base/limit pairs with a 64-byte
granule -- the Armv8-M model -- and the attribute encodings are bit-identical. `cortex_r4` is
PMSAv7, and its power-of-two region juggling does not apply. What differs and is handled: 20
EL1 regions rather than 8, `DFSR`/`DFAR` and `IFSR`/`IFAR` rather than CFSR/MMFAR, and no
secure-stack extension.

| Regions | Owner |
|---|---|
| 0-7 | Kernel / BSP boot table, never rebuilt by a module switch |
| 8-15 | Module block: kernel entry, code, data, five shared entries |
| 16 | The kernel's window onto the module area |
| 17-19 | Free on the S32Z280 |

The port needs at least 17 EL1 regions, so **it will not run on a Cortex-R52 configured with
16** -- a legal configuration. `mpu_init` refuses to enable protection rather than program a
region that does not exist.

Four divergences from the other module ports, each argued at its definition:

- `TXM_MODULE_MANAGER_REQUIRED_OPTIONS` is `USER_MODE | MEMORY_PROTECTION`, where every other
  port uses `0`. Any other combination leaves the region table zeroed and closes region 16, so
  the module cannot fetch its own first instruction.
- PMSAv8-R has no region priority, and an access hitting two enabled regions is an architected
  translation fault (TRM 8.1). The kernel's window is therefore enabled by thread ownership
  rather than bracketed around loader calls.
- A separate `threadx_module` kernel library, because the module `tx_port.h` changes four
  object layouts and two halves compiled against different headers link silently.
- The module ships as its own ELF pulled in with `.incbin`, built `-fpic -mno-long-calls` with
  a real GOT and `.data` rebase in `txm_module_gcc_setup.S`.

## Three fixes that reach past this port

- **The CLZ priority search had never run under GCC.** `tx_port.h` guarded it with
  `__TARGET_ARCH_ARM > 4`, an Arm Compiler 5 predefine, so the test read `0 > 4` and the
  portable loop ran instead -- in the default 32-priority configuration, not a dormant path.
  Now `__ARM_FEATURE_CLZ`. The same guard sits in about two dozen `tx_port.h` files tree-wide;
  the cross-port change is #650.
- **`txm_module_manager_external_memory_enable.c` did not mask the `PRLAR` limit** to the
  granule before ORing attributes, so `AttrIndx` came out 7 and selected an unwritten `MAIR`
  byte. Only this port's copy was audited; the same line is worth checking elsewhere.
- **The abort vector owes the fault handler a system-state bracket**, or `_tx_thread_terminate`
  never returns and the notify callback is unreachable. That is true of every AArch32 module
  port: `cortex_a7` does it correctly, `cortex_m33` gets away with it because its
  `_tx_thread_system_return` only pends PendSV.

## Validation

GNU Arm 14.3.1, 26 images across three configurations, zero warnings.

| Target | Result |
|---|---|
| FVP, default | `ctest` 8/8 |
| FVP, VFP/FIQ/nesting | `ctest` 11/11 |
| S32Z280-594EVB silicon | `s32z280_module.elf` PASS, `failures = 0`; `s32z280_demo.elf` PASS |

The model reports 32 EL1 regions, which no real R52 does, so a green `ctest` proves the logic
and not the region budget -- hence the board run. Every negative case is checked negatively:
grant the module one granule too many, or point it at an address it may touch, and the run
reports `MODULE RESULT: FAILED` and exits non-zero.

## Open items

#668 nothing in CI executes this suite, or any port. #669 regions 16-24 have direct `op1 = 1`
encodings that could replace the two `PRSELR` sequences, pending a proving run -- the TRM
contradicts itself on whether they exist. #670 Thumb-state modules are unvalidated. #650 and
#651 are the cross-port CLZ guard and the `const`-correctness of `tx_api.h` name parameters.

Documentation extends the R52 hardware-support page in `rtos-docs-asciidoc`.


fdesbiens:
Before asking anyone else to spend time on this, I ran my own review of the whole PR with a second agent — **Codex (GPT 5.6 Sol)** — against the PR as it stood on 26 August, covering all 52 changed files plus the base port they sit on, checked against the Cortex-R52 r1p3 TRM, the MP040 errata, the S32Z2 reference manual Rev. 5 and its mask-set errata. It found eight things. Four were merge blockers. This comment records what changed, and what deliberately did not.

## Fixed

**The module property contract could accept combinations that cannot run** (`ebe5a514`). `TXM_MODULE_MANAGER_REQUIRED_OPTIONS` was `0` while the port advertised user mode, memory protection and shared memory as supported, so the portable loader accepted an unprotected module and a user-mode-only module. The port only programs MPU entries when both bits are set, and the scheduler treated any module with a non-null data start as protected — it programmed a zeroed table and closed region 16, the kernel's only window over module memory. Those modules could not execute a single instruction. `REQUIRED_OPTIONS` is now `TXM_MODULE_USER_MODE | TXM_MODULE_MEMORY_PROTECTION`, and the scheduler reads the property flags itself rather than trusting a loader invariant it cannot see, so relaxing the contract later degrades to "runs under the kernel window" instead of faulting on the first instruction. A property-combination regression covers it.

**Shared-memory grants accepted a zero length and a wrapping end address** (`7f7d083b`). Both produce a limit below the base, and both were programmed into PRLAR, returned `TX_SUCCESS`, and consumed one of the five shared entries for a region the hardware cannot honour. Both are now refused with `TX_SIZE_ERROR` before anything is written or counted. The wrap test is written against the room left above the base rather than as `(address + length) < address`, which would have to commit the overflow to detect it.

**Thumb module entry points started with the wrong CPSR state** (`26186595`). The stack builder tested `r0` — `thread_ptr`, which is aligned — instead of the entry pointer, so the T bit was never set. The examples are forced to `-marm`, so every green run avoided the path. It now tests the entry pointer before that register is repurposed for CPSR. Thumb modules remain **documented as unvalidated** rather than claimed: there is no mixed A32-manager / Thumb-module image in the suite yet.

**An access matching two enabled regions was described as CONSTRAINED UNPREDICTABLE** (`9022695d`). TRM §8.1 specifies it as a translation fault. The abort observed on the S32 part is therefore the architected result, not a part-specific resolution of unpredictable behaviour, and the comments said otherwise in several places. Corrected. The mutual-exclusion design between the kernel window and module-owned regions is unchanged and still correct.

## Deliberately not changed, and why

**The granule round-up on shared grants.** A non-granule-multiple length maps up to 63 bytes beyond the request, because PRLAR describes an inclusive end of `LIMIT:0x3F`. Rejecting non-multiples was considered and rejected: `cortex_m33`'s `txm_module_manager_external_memory_enable.c` computes the same rounded grant and does not even mask the limit, so a stricter R52 would refuse working code moved from an Armv8-M part. The contract is now stated explicitly at the call site and in the documentation instead — including the deliberate asymmetry that `_txm_module_manager_inside_data_check()` validates against the *unrounded* length, so a module can reach the surplus directly while a kernel call carrying a pointer into it is refused. Pass a granule multiple and the two agree.

**Region 16 is still programmed through `PRSELR`.** The review was right that TRM §8.4 and Table 8-9 give `PRBAR16`/`PRLAR16` direct encodings at CP15 `op1 = 1`. But §3.3.85 and §3.3.86 contradict that and say direct access stops at `PRBAR15`. The selector sequence is the one proven on silicon, so swapping it for an encoding drawn from the half of the manual that disagrees with the other half is not something to do in the same change that owes a board run. The wrong comments were corrected; the optimisation belongs in its own change with its own before/after measurement on both targets.

## Validation at this head

Two FVP configurations under `ctest`: default **8/8**, VFP/FIQ/nesting **11/11**. All **26 images** across the three configurations build with GNU Arm **14.3.1** and zero warnings.

On S32Z280 silicon, the two images that link the kernel — and therefore the only two the CLZ priority-search change can reach — both ran and both passed:

| Image | Result |
|---|---|
| `s32z280_module.elf` | 7 PASS, 0 FAIL, `failures = 0` read back from the target |
| `s32z280_demo.elf` | `PASS threads, tick and preemption all verified` |

`s32z280_boot.elf` and `s32z280_atcm.elf` link `bsp_boot.c` and no kernel code — zero `clz` instructions in either — so they are build-only and no CLZ change can reach them.

Worth noting as a cross-check: the board reported a region-switch cost of **444–470 cycles** against the **434–470** this PR's description claims for the direct encoding. The silicon agrees with the text.

## Related

The documentation went in as eclipse-threadx/rtos-docs-asciidoc#41, now merged, and carries the corrected property, grant-length and register-access contracts so the two agree.

The R52 suite is not in CI today — nothing in CI executes any port — which is #688, separate from this PR. When both land, this PR's module and S32Z280 lanes belong in that workflow.

fdesbiens:
@billlamiework Let me know if you still intend to review this, or at least provide high-level feedback. Otherwise, I intend to merge it next week. Thanks!

PR #674, opened 2026-08-26
ports: add ARMv7-A SMP Linux build scripts(A5,A7,A9)

### LINKED ISSUE #605: Add .sh example build scripts for the ARMv7-A SMP ports (A5, A7, A9)
`scripts/check_clang.sh` links every example build it can drive, and reports the
ones it cannot. Three ports remain on that report:

```
no script driver, so outside this stage: cortex_a5_smp cortex_a7_smp cortex_a9_smp
```

Closing this empties that line. A check that permanently prints known exceptions
teaches people to skim past it, so zero is worth reaching.

## What is missing

Nothing much, which is what makes this a good first issue. All three examples are
already complete — `startup.S`, `sample_threadx.ld`, `sample_threadx.c`, and the
`MP_GIC` / `MP_Mutexes` / `MP_SCU` / `MP_PrivateTimer` / `v7` drivers are all
present under `ports_smp/<core>/gnu/example_build/`. They also already have
`build_threadx.bat` and `build_threadx_sample.bat`, which name every source and
every flag.

What they lack is the `.sh` counterparts, so nothing on Linux or in CI can build
them:

- `ports_smp/cortex_a5_smp/gnu/example_build/`
- `ports_smp/cortex_a7_smp/gnu/example_build/`
- `ports_smp/cortex_a9_smp/gnu/example_build/`

## How to do it

Follow the pattern #594 established for the 19 script-driven ports and #597
extended to the AArch64 SMP ports. Copy a `build_threadx.sh` /
`build_threadx_sample.sh` pair from a port that already has one, then translate
the two `.bat` files beside it: the library script is 257 lines of kernel sources,
and the sample script is only 7 or 8 lines.

Keep the `TOOLCHAIN=gnu|atfe` selection the existing scripts use, so
`check_clang.sh` can drive them with Arm Toolchain for Embedded as it does the
others.

That selection is also why this is worth a little more than the LLVM check alone
suggests. There is no GCC equivalent of `check_clang.sh` today — GCC cross-compiles
nothing in CI beyond the four Cortex-M ports — and adding one is planned. Because
these scripts default to `TOOLCHAIN=gnu`, the same pair serves both checks: whoever
writes the GCC side gets these three ports for free rather than reporting them as a
second permanent exception. So please do not hard-code a compiler; take it from the
`TOOLCHAIN` case the other scripts use.

## Three wrinkles worth knowing before you start

These are the things that will cost you an hour if nobody tells you:

1. **The `.bat` files emit `sample_threadx.axf`,** but `check_clang.sh` decides
   whether an example built by looking for `sample_threadx.out`. Emit `.out`, as
   the other ports' scripts do, or the stage will report a failure for an example
   that linked fine.
2. **They link `-e Vectors`,** not the `reset_handler` entry symbol the Cortex-M
   and Cortex-A examples use. Keep `Vectors`; it is what `startup.S` defines here.
3. **`cortex_a5_smp` has `tx_initialize_low_level.s` in lowercase.** A lowercase
   `.s` gets no preprocessing from the compiler driver. Check whether that file
   needs any before renaming it, and if it does not, leave the name alone and just
   reference it correctly.

## How to check your work

```
scripts/check_clang.sh --clang /path/to/clang
```

The three cores should disappear from the "no script driver" line, and the
script-driven example count should rise from 42 to 45. Please build with both
toolchains — plain GNU and `TOOLCHAIN=atfe` — since catching GNU-versus-LLVM
divergence is the reason that check exists.

## Scope note

These images are link-verified only. There is no Cortex-A9 SMP model in CI, so a
green result means "it links", not "it runs" — the same standard #597 set for the
AArch64 examples. Please do not claim more than that in the PR description.

The ports' own kernel sources are already covered by the assembly stage of the
same script, so this issue is about the examples rather than the ports.

prashit-vora:
## Summary

- Add GNU/ATfE build scripts for the Cortex-A5, Cortex-A7, and Cortex-A9 SMP libraries and examples.
- Handle the lowercase assembly sources that require preprocessing on Linux.
- Add the corresponding local `crt0.S` to each example and link with `-nostartfiles`, keeping `Vectors` as the entry point.

Closes #605

## Verification

- GNU Arm Embedded GCC 13.2.1: `build_threadx.sh` and `build_threadx_sample.sh` completed for A5, A7, and A9; all three sample images linked.
- ATfE 22.1.0: `scripts/check_clang.sh` completed with 724/724 assembly sources, all feature-macro checks, all common C profiles, and 45/45 example builds passing.
- CMake example stage: Cortex-R52 linked 5/5 images.

Relevant output from `check_clang.sh`:

```text
== Assembly sources of every Arm gnu port ==
  724 of 724 assembled

== Common C sources, one core per architecture profile ==
  cortex_m0: 185 of 185 compiled
  cortex_m4: 185 of 185 compiled
  cortex_m23: 185 of 185 compiled
  cortex_m33: 185 of 185 compiled
  cortex_m55: 185 of 185 compiled
  cortex_a7: 185 of 185 compiled
  cortex_a53: 185 of 185 compiled
  cortex_r5: 185 of 185 compiled
  cortex_r52: 185 of 185 compiled

== Example builds, linked with lld ==
  45 of 45 example builds linked

== CMake example builds, linked with lld ==
  cortex_r52: 5 of 5 images linked

All LLVM toolchain checks passed.
```

NOTE: These examples were link-verified only and were not run on hardware.

### DIFF
--- a/ports_smp/cortex_a5_smp/gnu/example_build/build_threadx.sh
+++ b/ports_smp/cortex_a5_smp/gnu/example_build/build_threadx.sh
@@ -37,7 +37,7 @@ case "${TOOLCHAIN}" in
 esac
 
 rm -f tx.a
-"${CC}" ${TARGET_FLAGS} -c -g -mcpu=cortex-a5 -x assembler-with-cpp tx_initialize_low_level.s
+"${CC}" ${TARGET_FLAGS} -c -g -mcpu=cortex-a5 tx_initialize_low_level.S
 "${CC}" ${TARGET_FLAGS} -c -g -mcpu=cortex-a5 ../src/tx_thread_context_restore.S
 "${CC}" ${TARGET_FLAGS} -c -g -mcpu=cortex-a5 ../src/tx_thread_context_save.S
 "${CC}" ${TARGET_FLAGS} -c -g -mcpu=cortex-a5 ../src/tx_thread_interrupt_control.S
--- a/ports_smp/cortex_a5_smp/gnu/example_build/build_threadx.sh
+++ b/ports_smp/cortex_a5_smp/gnu/example_build/build_threadx.sh
@@ -1,259 +1,281 @@
 #!/bin/sh
 set -e
+
+# this script builds the ThreadX library for the Cortex-A5 SMP port using
+# GNU toolchain or llvm/clang toolchain (defaults to GNU).
+: "${TOOLCHAIN:=gnu}"
+
+case "${TOOLCHAIN}" in
+    gnu)
+        CC="arm-none-eabi-gcc"
+        AR="arm-none-eabi-ar"
+        TARGET_FLAGS=""
+        ;;
+    atfe)
+        CC="${ATFE_CLANG:-clang}"
+        AR="llvm-ar"
+        TARGET_FLAGS="--target=arm-none-eabi"
+        ;;
+    *)
+        echo "Unknown TOOLCHAIN: ${TOOLCHAIN}" >&2
+        exit 1
+        ;;
+esac
+
 rm -f tx.a
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 -x assembler-with-cpp tx_initialize_low_level.s
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_context_restore.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_context_save.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_interrupt_control.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_interrupt_disable.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_interrupt_restore.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_irq_nesting_end.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_irq_nesting_start.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_schedule.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_core_get.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_core_preempt.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_current_state_get.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_current_thread_get.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_initialize_wait.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_low_level_initialize.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_protect.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_time_get.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_smp_unprotect.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_stack_build.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_system_return.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_thread_vectored_context_save.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 ../src/tx_timer_interrupt.S
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 -I../../../../common_smp/inc -I../inc ../../../../common_smp/src/tx_block_allocate.c
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 -I../../../../common_smp/inc -I../inc ../../../../common_smp/src/tx_block_pool_cleanup.c
-arm-none-eabi-gcc -c -g -mcpu=cortex-a5 -I../../../../common_smp/inc -I../inc ../../../../common_smp/src/tx_block_pool_create.c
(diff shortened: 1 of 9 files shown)


fdesbiens:
Thanks for this — these are exactly the three ports `scripts/check_gcc.sh` reported as having no script driver, so it takes the GCC example-build stage from 42 to 45.

I've **rebased your branch onto current `dev` and pushed one small commit on top of yours**, so it can go through the check that was written to cover it. Your eight commits are unchanged apart from the new base; the extra commit is two lines. Details below so nothing is a surprise.

**Heads up on your local clone**: the rebase rewrote the commit SHAs, so before you touch the branch again you'll want

```
git fetch origin && git reset --hard origin/armv7-smp-linux-build-scripts
```

rather than pulling, or git will try to merge the old and new histories together.

---

### What the extra commit changes, and why

Two lines used `-x assembler-with-cpp` to preprocess a lowercase `.s` source. That was the right call when you wrote it. In the meantime #672 landed, and it fixes the underlying problem by renaming rather than by working around it:

| file | was | now |
|---|---|---|
| `cortex_a5_smp/…/build_threadx.sh` | `-x assembler-with-cpp tx_initialize_low_level.s` | `tx_initialize_low_level.S` |
| `cortex_a7_smp/…/build_threadx.sh` | `-x assembler-with-cpp ../src/tx_thread_smp_unprotect.s` | `../src/tx_thread_smp_unprotect.S` |

The reason for the rename is that GAS runs the C preprocessor on `.S` and not on `.s`. In a `.s` file every `#` line is a comment, so a `#define` is never substituted and an `#if`/`#else` pair emits **both** arms. Twenty-nine files under `gnu` trees were on the wrong side of that and only three of them failed to assemble — the rest just quietly did the wrong thing. The worst was `ports_smp/cortex_a7_smp/gnu/src/tx_thread_smp_unprotect.s`, one of the two files your scripts name: it ignored all four of its own feature macros, wrote the caller's LR into the protection structure on every unprotect, sent an unconditional `SEV`, disabled FIQ regardless of `TX_ENABLE_FIQ_SUPPORT`, and returned twice.

So after #672 the extension carries the preprocessing and no caller has to remember the flag — and the lowercase names no longer resolve at all. Unmodified on the new base, `cortex_a5_smp/build_threadx.sh` failed with `cc1: fatal error: tx_initialize_low_level.s: No such file or directory`. Those were the only two lines affected; the other seven references to the same two files across your three scripts already named them with a capital `S`.

The rebase itself was clean — no conflicts, and the nine files you add are untouched.

### Verified rather than assumed

With the pinned Arm GNU 14.3.rel1 toolchain, on the rebased branch:

- `build_threadx.sh` and `build_threadx_sample.sh` both succeed for a5, a7 and a9, and all three link a `sample_threadx.out`.
- `scripts/check_gcc.sh` passes end to end — 840 of 840 assembly sources, 185 C sources across nine core profiles, and **45 of 45 example builds linked**, up from 42.
- `scripts/check_clang.sh` (ATfE 22.1.0) agrees at **45 of 45**.
- `scripts/check_ports.sh` passes, including its new section asserting that no `.s` file under a `gnu` tree uses the preprocessor.

Your `crt0.S` with `-nostartfiles` is the same shape as a defect #673 has just fixed for the AArch64 samples, where `-nostartfiles` was paired with `--specs=rdimon.specs` and newlib's startup then referenced `_init`/`_fini` that nothing linked. You are using `--specs=nosys.specs`, which never reaches `__libc_init_array`, and the three links above confirm these ports are unaffected. Good instinct.

### Two things about the checks, neither of them your fault

**Nothing had actually run on this PR before now.** The checks list showed only `eclipsefdn/eca`, but `clang_check`, `ports_arch_check` and `regression_test` all had runs sitting at `action_required` — GitHub holds workflow runs on fork pull requests until a maintainer approves them, and nobody had. That is on us. They are approved now, and the hold comes back on every push, so ping us if the list ever looks empty again.

**`gcc_check / gnu` could not have run on the old base at all.** For a `pull_request` event GitHub decides which workflows run from the *head* branch's copy of the workflow files, and that workflow only landed on `dev` this morning in #675. Rebasing is the only thing that makes it appear — closing and reopening does not. It has now run, and passed in 4m37s, so these three example builds are gated by the check written for them rather than merged past it. Its log reads `840 of 840 assembled` and `45 of 45 example builds linked`, against 42 before this PR — matching what I measured locally, and with the pinned 14.3.rel1 on both drivers.

One last note, purely informational: your verification used GCC 13.2.1, which is what the distro `arm-none-eabi-gcc` package usually is. The project pins Arm GNU 14.3.rel1, and it is an easy one to pick up without noticing — everything above was re-run against 14.3.rel1. In the same vein, the `check_clang.sh` output in the description reports 724 of 724 assembly sources; that number is 840 after #672, which corrected a glob that had been silently matching nothing, so 116 files across nine module ports had been assembled by no check at all. Nothing wrong on your side in either case.

All seven checks are green and the PR is ready to merge. Please do give the extra commit a look first, and shout if you'd rather it were done differently.

prashit-vora:
Thanks for updating the branch and explaining everything. I checked the extra commit and understand the .S changes after #672. I also noted the GCC 14.3 version and the updated assembly count of 840. Thanks for rerunning the checks.

fdesbiens:
Re-verified against current `dev` before merging, since this has been open a while and three of the twenty-four commits since touched these exact ports — #704 (missing memory barrier in the ARMv7-A SMP schedulers), #707 (removed FIFO queueing, which deleted six `tx_thread_smp_protection_wait_list_macros.h` files) and #713 (include directories marked SYSTEM).

Nothing needs changing. It merges cleanly, and no script references a file #707 removed.

With the pinned Arm GNU 14.3.rel1, on the merged tree:

| Check | Result |
| --- | --- |
| `build_threadx.sh` + `build_threadx_sample.sh`, A5 / A7 / A9 | all six succeed, all three link |
| `scripts/check_gcc.sh` | passes — **45 of 45** example builds linked |
| `scripts/check_ports.sh` | passes, including the assertion that no `.s` under a `gnu` tree uses the preprocessor |

Two things I checked that were not covered before:

**Every port source is built.** Each script names 21 of the 21 sources in its port's `gnu/src`, so nothing is silently left out — worth knowing because a source added later would otherwise slip past these scripts without anyone noticing.

**The linked images are sound.** All three report entry point `Vectors` at `0x80008000` and, more to the point, **zero undefined symbols**. That settles the `crt0.S` question raised earlier: `-nostartfiles` with `--specs=nosys.specs` genuinely never reaches `__libc_init_array`, so these ports do not have the `_init`/`_fini` defect #673 fixed for the AArch64 samples.

One caveat, not a reason to hold this up: the images are link-verified only and nothing in CI executes ARMv7-A SMP. Both #704 and #707 were runtime defects in these ports — a missed preemption and an ISR deadlock — and neither is the kind of thing a link can catch. That gap is the same one #688 has just started closing for the Cortex-R52, and these three ports would be the natural next candidates once there is a way to run them.

Thanks again @prashit-vora.

PR #676, opened 2026-08-28
Stopped the win32 and win64 ports from enabling performance metrics and event trace

ntfreak:
if tracing or performance counters are required they can be set by the user in tx_user.h

## PR checklist


fdesbiens:
Thanks @ntfreak — this is the right change, and the justification is stronger than the description gives it.

Checking every port's `tx_port.h` with comments stripped, these two were the **only** ports that forced the symbols on:

| Symbol | Ports defining it before this PR |
| --- | --- |
| `TX_*_ENABLE_PERFORMANCE_INFO` | 2 — win32, win64 |
| `TX_ENABLE_EVENT_TRACE` | 1 — win32 |

So this does not introduce a new convention, it removes the last exceptions to the existing one. With it in, no port defines either. That matters more than a build-size preference: the symbols extend the control blocks, so a port setting them behind the application's back changes structures the application also sees.

(The linux port carries the same list, but commented out — it reads at a glance like a third case and is not one. Worth knowing before someone "finishes the job" there.)

### What I pushed

**The disabled blocks are removed rather than `#if 0`-ed.** `#if 0` is the form MISRA C:2012 Directive 4.4 is about — sections of code should not be commented out — and it left two copies of a list with no reader. A short note in their place records where the symbols belong and why the port does not set them. Preprocessor nesting in both headers re-checked as balanced.

No behaviour change beyond what you already had.

### Retargeted to `dev`

The base was `master`. Beyond the branching convention, it has a concrete cost: `clang_check` and `gcc_check` exist only on `dev`, and a pull request runs the workflows from its **base** branch, so targeting `master` skipped both. `dev` is where this work belongs and where the checks live. The two files are byte-identical on the two branches, so the retarget was clean.

### On the CI that did not run

For the record, since it looked like a gap: the runs were not missing, they were held at `action_required` awaiting maintainer approval, which is standard for an outside contributor. A maintainer approving them releases `ports_arch_check` and `regression_test`.

It is worth being honest that approving them proves little here. **There is no Windows CI in this repository** — no workflow references win32, win64, MSVC or a Windows runner — so nothing in CI compiles either file, before or after this change. That is not something to solve in this pull request, but it is why this one is being read carefully rather than waved through on a green tick.

PR #691, opened 2026-09-03
riscv32: spec compliance and regression test fix

akifejaz:
This PR corrects the RISC-V32 GNU port as per spec and its QEMU virt support.

- The port uses 16-byte stack alignment at each C function call.
- Context paths preserve integer, floating-point, CSR, and program-counter state.
- Frame offsets use the `slot*byte-size(base)` format.
- Build guards reject unsupported ISA and ABI combinations.
- Startup code initializes `gp` and uses valid linker regions.
- CLINT code uses safe RV32 timer reads and writes.
- PLIC code masks unused sources and handles claim ID zero.
- The QEMU test checks timer, FPU, PC, time-slice, clock, and preemption behavior.


akifejaz:
This and the following few PRs will be related to RISC-V CI enablement efforts. I'had discussed this in last Weekly meeting.

I've started testing the riscv ports with available tests and closely comparing implementations with respect to riscv spec. 

Once the CI of riscv ports is running cleanly, I'll be processing with Cloud-V setup (RISC-V hardware CI)

cc @fdesbiens

akifejaz:
@fdesbiens this PR was supposed to be merge before the #698, but anyways. Let me know if you face any conflicts or other issues.

fdesbiens:
Thanks for your patience on this one, @akifejaz — and apologies for the ordering. This was meant to land before the RISC-V64 work, and merging that first is what caused the detour. That was on me, not on you.

I have merged `dev` into your branch and pushed two small follow-ups. The rest of the PR stands as you wrote it.

### Verified

With the RISC-V64 fixes now in and the test runner no longer masking incomplete builds, the whole RISC-V regression suite is green on both architectures:

| Suite | Result |
| --- | --- |
| RISC-V32, 5 configurations | **475/475**, nothing unlinkable |
| RISC-V64, 5 configurations | **475/475**, nothing unlinkable |

Both `check-functional-riscv32` configurations pass every check — hard float with `FPU_VERIFIED_OK`, soft float with the skip path, and `MEPC_VERIFIED_OK`, time-slice, system timer and preemption in both.

Worth calling out: your `time()` implementation in `syscalls.c` is what finally clears `threadx_thread_delayed_suspension_test` and `threadx_thread_wait_abort_and_isr_test`. Those two could not link at all before — newlib pulled `_impure_ptr` out of `R_RISCV_HI20` range — and they failed on RISC-V64 as well, because the BSP is shared. Your change fixes both ports.

### What I merged

`test/tx/cmake/riscv/bsp/tx_initialize_low_level.S` conflicted with #708, which had already taught that file to take the frame sizes from the port for both XLENs. I took the `dev` side, so your `__riscv_xlen == 32` guard is gone as redundant. Your `MIE_MSIE` removal in the same file is kept.

### What I changed

**The frame sizes now come from one place.** `tx_port.h` published `TX_RISCV_TRAP_FRAME_SIZE`, but nothing in the port consumed it — six `.S` files each rebuilt the same numbers from their own `#if`, so the interrupt frame size appeared in seven places and the solicited frame size in three. That is exactly the shape that caused the RISC-V64 fault in #708: the port moved to a padded frame and one copy of the constant did not follow. The sources now include `tx_port.h` and take both sizes from it, and `TX_RISCV_SOL_FRAME_SIZE` joins the contract since the solicited frame was never published. Emitted code is unchanged — 400 and 176 for ILP32D, 128 for soft float, confirmed by disassembly on both sides.

**`_tx_initialize_low_level` had `.global` directly followed by `.weak`**, so the symbol stayed weak and the `.global` did nothing (`nm` shows `W`). Weak is what the port wants, since the example and regression BSPs both supply their own definition, so I removed the stray `.global` rather than the `.weak`.

**The QEMU runner seeded `fpu_verified` from `skip_fpu`**, so a soft-float run satisfied the FPU gate whether or not the script ever reported the skip. It now starts false and is set only when the marker appears, so a run that dies early fails instead of passing.

Nothing needed from you. Merging once CI is green.

PR #698, opened 2026-09-07
riscv64: spec compliance and regression test fix

akifejaz:
This PR corrects the RISC-V64 GNU port as per spec and its QEMU virt support.

- Fix RISC-V64 GNU port and QEMU virt BSP for spec compliance and proper FP/vector context saving.
- Enforce atomic interrupt state operations, correct stack frame layouts, and add compile-time ABI guards.
- Refactor PLIC and CLINT drivers with volatile memory accesses, proper indexing, and exception skip logic.
- Update linker scripts, global pointer initialization, and CMake functional test targets.

Regression 475/475 on QEMU 6.2 (5 configs x 95), compile matrix for lp64d, lp64f, lp64 and rv64gcv, ecall/ebreak skip checked on RV64 and RV32.


fdesbiens:
Thank you for this, @akifejaz — it is a thorough piece of work, and the parts that matter most are the parts that are easiest to get wrong. The atomic `csrrc`/`csrrs` rewrite of `_tx_thread_interrupt_control`, the `bgtz` → `bnez` corrections on `_tx_thread_preempt_disable` and `_tx_timer_time_slice` (both of which silently mishandled a wrapped or negative value), and removing the duplicate vector save in `_tx_thread_preempt_restore` are all real bugs caught and correctly reasoned about.

Because the RISC-V jobs are still commented out in `regression_test.yml`, the green checks on this PR say nothing about this port, so I built and ran it locally.

### One blocking issue, now fixed

The new `check-functional-riscv64` target failed on every run:

```
FPU_VERIFIED_FAIL_NO_SYMBOL
FAILURE: Hit thread, but this demo defines no fpu_test_val, so the FPU result is unverifiable.
```

Commit 0e95233 added `fpu_test_val` to `demo_threadx.c`; 3a374f4a ("revert the demo changes") took it back out four minutes later. The GDB script deliberately treats the missing symbol as a failure rather than skipping the check — which is the right call — so the target could not pass with the demo as it stood. I suspect the functional run quoted in the description predates that revert; the 475/475 figure is the `test/tx` suite, which does not exercise this path.

I have restored those three lines, matching what the risc-v32 demo already carries. The target now passes end to end:

```
FPU_VERIFIED_OK value=1.100000 adds=1
MEPC_VERIFIED_OK mepc=0x80000720 frame_pc=0x80000720
SUCCESS: Time-slice handler verified.
SUCCESS: System timer increment verified.
PREEMPT_VERIFIED_OK ticks=1
```

### Three small things folded into the same commit

- `tx_port.h` carried a comment saying the ISA string must include Zicsr, but nothing enforced it, so an `rv64imac` build produced a wall of assembler `unrecognized opcode` errors. It now stops at one clear `#error`.
- `TX_RISCV_TRAP_CALL_FRAME_SIZE` was defined and never referenced; removed.
- The example `.gitignore` listed `qemu-riscv32.log`, but the runner writes `qemu-riscv64.log`.

### Verified

I confirmed the ABI matrix from your description by assembling every port source:

| Configuration | Result |
| --- | --- |
| `rv64gc` / `lp64d` | builds |
| `rv64gc` / `lp64f` | builds |
| `rv64imac_zicsr` / `lp64` | builds |
| `rv64gcv` / `lp64d` | builds |
| `rv64gc` / `lp64` | rejected by the soft-float guard, as documented |
| `rv64imac` / `lp64` | rejected by the new Zicsr guard |

I also checked the frame-size unification for self-consistency across save, restore, schedule, stack build and system return (528 = 65×8 plus 8 bytes of padding with FP, 256 without; 240/128 for the solicited frame), and confirmed the branch still merges cleanly onto `dev` with the `ra` fix from #696 intact.

Merging shortly. Thanks again for the care that went into this one.

fdesbiens:
A correction on my part, @akifejaz, and an apology.

The commit I pushed to your branch (3b6fc4f0) carried nine files it should not have. I had configured a local CMake build tree while verifying the functional test, and it was staged along with the fix — a compiled `kernel.elf`, `build.ninja`, the Ninja dependency and build logs, a QEMU run log, and some CPack scaffolding. Roughly 830 KB of generated output, none of it yours and none of it part of this port fix.

That is entirely my mistake, not a defect in your contribution. The RISC-V64 work itself is correct and stays as merged.

#706 removes the files from `dev`. It also replaces the `.gitignore` list of build directory names with a single `build*/` pattern, which is what would have caught this in the first place.

Nothing is required from you here. Thanks again for the contribution, and sorry for the noise on your PR.

PR #717, opened 2026-09-10
Enabled the RISC-V regression suite in CI for QEMU targets

akifejaz:
This PR enables the RISC-V CI tests. It runs about 5 configurations for RV64, RV32 each with around 955 tests combined.

These use QEMU for running RV32 and RV64 tests.

### DIFF
--- a/.github/workflows/regression_test.yml
+++ b/.github/workflows/regression_test.yml
@@ -113,9 +113,21 @@ jobs:
       cmake_path: ./test/tx/cmake/riscv
       result_affix: RISC-V
       skip_deploy: true
-      # No coverage build configuration for the RISC-V suite yet: the CMake
-      # configurations under test/tx/cmake/riscv match the Linux suite's minus
-      # the coverage instrumentation, so gcovr has nothing to read here.
+      # No coverage from this suite, and not for want of a build configuration.
+      # These tests are bare-metal images run under QEMU with -bios none, and
+      # test/tx/cmake/riscv/bsp/syscalls.c gives them _write to the UART, _exit
+      # through the sifive_test device, and stubs for _close, _fstat, _isatty,
+      # _lseek, _read and _sbrk -- but no _open. gcov emits a .gcda by opening
+      # a path, so instrumenting these builds would produce nothing to read.
+      #
+      # test/tx/cmake/riscv/CMakeLists.txt is its own top-level project and does
+      # not declare the TX_COVERAGE option that test/tx/cmake and test/smp/cmake
+      # do, so the OFF the template exports is inert here rather than the thing
+      # holding coverage off.
+      #
+      # Collecting it therefore needs a transport off the target -- gcov's dump
+      # routines over the UART the BSP already drives, or semihosting -- and not
+      # a default_build_coverage entry beside the other five configurations.
       skip_coverage: true
   deploy:
     # Publishing the coverage report stays on master. Adding dev to this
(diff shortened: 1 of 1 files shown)


fdesbiens:
Thanks for this, @akifejaz — this is a good change and the enablement is clean. I reviewed it and pushed one comment-only commit to your branch (ce1ea7b); the job's inputs are untouched.

## What I checked

The wiring is right. `run.sh` writes its JUnit report to `build/<arch>_<build>/<arch>_<build>.xml`, which is what the template's `${cmake_path}/build/*/*.xml` glob expects, and both the check name (`Test Results RISC-V`) and the artifact name are distinct from the other three suites. Leaving `deploy`'s `needs: [tx, smp]` alone is correct, since this suite publishes no coverage report.

I also confirmed the run is real rather than an empty pass — the numbers in your description hold up:

| | configurations | tests each | total |
|---|---|---|---|
| RV32 | 5 | 95 | 475 |
| RV64 | 5 | 96 | 480 |

955, all green. RV64's extra one is `threadx_riscv_new_thread_fpu_state_test`, added conditionally and already explained in the CMake. And the suite is very close to complete against the Linux one — only `threadx_thread_misaligned_stack_test` and `threadx_trace_entry_update_test` are absent, both already reasoned about in `test/tx/cmake/riscv/regression/CMakeLists.txt`.

It is also cheap, which matters for something that now runs on every pull request: install 97s, build 88s for all ten configurations, test 111s. That is roughly 6x headroom against the step timeouts.

## The one tweak

I rewrote the `skip_coverage` comment. What you wrote is accurate as far as it goes — the configurations under `test/tx/cmake/riscv` are the Linux suite's minus the coverage one, and gcovr has nothing to read. The word I changed is "yet", because it points the next reader at a fix that would not work.

Coverage here is not one missing `default_build_coverage` entry. These tests are bare-metal images run under QEMU with `-bios none`, and `test/tx/cmake/riscv/bsp/syscalls.c` gives them `_write` to the UART, `_exit` through the sifive_test device, and stubs for `_close`, `_fstat`, `_isatty`, `_lseek`, `_read` and `_sbrk` — but no `_open`. gcov emits a `.gcda` by opening a path, so instrumenting these builds produces nothing regardless of how they are configured. Nor is the template's `TX_COVERAGE=OFF` what holds coverage off: `test/tx/cmake/riscv` is its own top-level project and never declares that option, so the value is inert there.

Getting a figure out of this suite needs a transport off the target — gcov's dump routines over the UART the BSP already drives, or semihosting. In a project that asks for 100% coverage that is worth saying plainly, so nobody spends an afternoon adding a configuration that cannot help.

## At merge

I will retitle this to "Enabled the RISC-V regression suite in CI for QEMU targets" when it goes in. Merges here squash under the PR title, and commit subjects in this repository start with a past-tense verb — nothing for you to change.

## Follow-ups I will take separately

Neither is a change to this PR, and neither should hold it up:

- `scripts/install_riscv.sh` is unchanged here, but this PR is what puts it on every pull request's critical path, and it does not yet have the hardening `scripts/install.sh` grew for the same reason: it runs under `set -e` with a bare `apt-get update`, where `install.sh` deliberately downgrades a failed update to a warning after an unrelated third-party repository on the runner image turned builds red. Its two `wget` calls have no retry or timeout either. Today's run had ample headroom, so this is about flake resistance, not speed — I will lift the existing `retry`/`timeout` helpers across, and add checksums to the tarballs while I am in there. The release tag being pinned is the important half already.
- `gcc_check.yml` excludes RISC-V on the grounds that it is "a family that is not regressing". Still true, and better supported now that an executing suite covers it, but the sentence reads as though nothing in CI runs RISC-V, which stops being the case when this merges. I will refresh it.

Nice work — this closes a real gap.

